PR #186, opened 2013-04-07
Updates to Paned and TextCellView

Jeha:
I have implemented the following changes:
1) Xwt.Paned did not received MovePanel signals from GTK.Paned, so I watched the events being sent to Paned and detected any time that the panel had been moved, then manually fired the event to the EventSink.
2)  Updated the TextCellView to allow for font family, font weight, font size and alignment control.

This is my first time to ever try to make a change to a program's gitsource, so I hope this is correct.  Please advise me if my method and/or style is not correct.

Thanks.


sevoku:
I think the Paned part can be merged. The alignment needs an other review I think. I'll take a look on it next days.

sevoku:
I have cherry-picked the Paned part in #472 (commit hashes changed). So only the Alignment part remains.

PR #226, opened 2013-07-13
[Mac] fix color chooser sample

dlech:
Color chooser sample is broken on Mac. There are 2 separate problems. 

1. Color chooser uses XwtWidgetBackend instead of a native backend, so trying to cast the backend to a ViewBackend throws and exception. I added the method GetNativeBackend() to recursively get a ViewBackend object in the case of an XwtWidgetBackend object.

2. After fixing the above problem, there is no more exception thrown, but the color chooser is still not visible. It turns out you need to call SetAutosizeMode (true) in order for it to be visible, so I added code to set this for all XwtWidgetBackend objects.


sevoku:
After some testing and playing around it looks like setting SetAutosizeMode(true) is not the right way, since the child will always have the same size. I've merged your fix into  #446 and added direct resizing to CustomWidgetBackend. Works for the ColorChooser and the new FontChooser (#443).

PR #243, opened 2013-09-19
Implement treestore sorting

riviti:
This makes it possible to tell TreeStore to sort the rows by one of the columns. Implemented for GTK and WPF. Mac is impossible for me to do since I do not have a system to test on.

### DIFF
--- a/Xwt.WPF/Xwt.WPFBackend/TreeStoreBackend.cs
+++ b/Xwt.WPF/Xwt.WPFBackend/TreeStoreBackend.cs
@@ -183,7 +183,30 @@ namespace Xwt.WPFBackend
 			this.topNodes.Clear();
 		}
 
-		public IEnumerator GetEnumerator ()
+		#region Sorting handling
+
+		public void SetSortField(IDataField field, ColumnSortDirection order)
+		{
+			SortField = field;
+			SortOrder = order;
+			OnSortOptionsChanged(new SortingOptionsEventArgs(field, order));
+		}
+
+		public event EventHandler<SortingOptionsEventArgs> SortOptionsChanged;
+
+		protected virtual void OnSortOptionsChanged(SortingOptionsEventArgs e)
+		{
+			var handler = SortOptionsChanged;
+			if (handler != null) handler(this, e);
+		}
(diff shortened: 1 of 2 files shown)


Xwt.Gtk/Xwt.GtkBackend/TreeStoreBackend.cs
```diff
@@ -86,6 +86,11 @@ public void Clear ()
 			Tree.Clear ();
 		}
 
+		public void SetSortField(IDataField field, ColumnSortDirection direction)
+		{
+			Tree.SetSortColumnId (field.Index, direction == ColumnSortDirection.Ascending ? Gtk.SortType.Ascending : Gtk.SortType.Descending);
+		}
+
```
sevoku: This is redundant. Sorting can be set on the columns directly.

sevoku:
Any progress here? I'm going to look on sorting next weeks. Has someone tried this approach? Can the backend code be reused?

Xwt/Xwt/TreeStore.cs
```diff
@@ -69,6 +69,11 @@ public TreeStore (params IDataField[] fields)
 		ITreeStoreBackend Backend {
 			get { return (ITreeStoreBackend)BackendHost.Backend; }
 		}
+
+		public void SetSortField (IDataField field, ColumnSortDirection order)
```
sevoku: This is redundant. Sorting can be set on the columns directly.

riviti: As far as I remember, for the Gtk+ backend this is (was) not possible, because the Xwt column does not have access to the Gtk+ tree store, which is where sorting is controlled in that toolkit. I remember there was some Xwt api concern about this PR which is the reason for it not being merged, but the discussion doesn't seem to have happened here, perhaps it was in #mono or #monodevelop.

riviti: Well, I don't have any immediate interest in getting this merged (anymore), so since it seems to have bitrotted, I'd say whoever feels the need should feel free to pick this work up and finish it.

PR #266, opened 2013-11-24
Wpf removing item kept it in SelectedItems list

DavidKarlas:
Wpf when removing item from TreeView WpfBackend did not remove item from SelectedItems.

I tried not to modify TreeNodeChildEventArgs class but problem was that I could not identify removed node based on parent + childIndex because it was already removed from DataSource before NodeDeleted event was called.

To see this bug navigate to TreeViews sample in Samples and 2x click Remove selected item. First time it will remove properly second error will occure because SelectedItems still has removed item.

### DIFF
--- a/Xwt.WPF/Xwt.WPFBackend/TreeStoreBackend.cs
+++ b/Xwt.WPF/Xwt.WPFBackend/TreeStoreBackend.cs
@@ -155,7 +155,7 @@ namespace Xwt.WPFBackend
 			int index = list.IndexOf (node);
 			list.RemoveAt (index);
 
-			OnNodeDeleted (new TreeNodeChildEventArgs (node.Parent, index));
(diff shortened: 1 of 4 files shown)


sevoku:
This is really old and I can not reproduce this on the current master. Already fixed by some other PR, maybe? And the changes should be applied to all other backends, too. Please create a new PR or Issue, if this is still not fixed. I'll take a look at Wpf SelectedItems behavior next time I work on the tree implementations.

sevoku:
Ok, I assumed that some exception will be thrown. The Wpf backend does not update the SelectedItems list, when a selected node is removed. I've created a new issue: #452

The approach here is not really good, because when the event is (should be) triggered by the TreeStore, the item should be removed already, so the TreePosition would become invalid (even if it is contained in the event args, it is useless for the user.) So this should be fixed on the backend side, and only there.

PR #268, opened 2013-12-01
[gtk] Add call to BeginPaintRect()/EndPaintRect() before and after Event...

atsushieno:
...Sink.OnDraw().

This enables double buffering for user code.


sevoku:
I think this can be closed. If someone experiences the same problem, please open a new issue with some example Xwt based code, so it can be reproduced by others.

Issue #279, opened 2014-01-26
Wpf window spawns outside of the screen


MaikKlein:
I am using Windows 7 x64, Wpf and F# and the sample code spawns the window outside of my screen.

```
open System
open Xwt

[<STAThread>]
[<EntryPoint>]
let main argv = 
    Application.Initialize(ToolkitType.Wpf)
    let w = new Window (
               Title = "Xwt Demo Application",
               Width = 500.,
               Height = 500.,
               Location = Point(500.,500.)
    )
    w.Show()
    Application.Run()
```
It also completely ignores the height and width. As you can see I also tried the set the location manually but it also made no difference.

It seems to spawn somewhere around -100,-100

sevoku:
fixed via #393

PR #281, opened 2014-01-26
Initial window position without content fix for #279

DavidKarlas:
If window has Content everything is same except "} else if (locationSet && !shown)" where !shown is removed because it didn't make sense because it was always true but actually false(see code at start of method)

Atm InitialLocationManualWithoutContentAndSize is failing for WPF because there is some bug in WPFBackend.Move method compering to WPFBackend.Bounds(called when Size is set)
I think it's somehow connected to Xwt.WPFBackend.WpfWindow.initialX not being set by Move method.

All 3 methods are failing for GTK because known bug of 8px offset.

### DIFF
--- a/Testing/Tests/WindowTests.cs
+++ b/Testing/Tests/WindowTests.cs
@@ -218,6 +218,44 @@ namespace Xwt
 			}
 			Assert.IsFalse (closed, "Window should not be closed");
 		}
+
+		[Test]
+		public void InitialLocationManualWithoutContent ()
+		{
+			using (var win = new Window ()) {
+				win.InitialLocation = WindowLocation.Manual;
+				win.Location = new Point (210, 230);
+				win.Size = new Size (100, 100);
+				ShowWindow (win);
+				Assert.AreEqual (210, win.X);
+				Assert.AreEqual (230, win.Y);
+			}
+		}
+
+		[Test]
+		public void InitialLocationManualWithoutContentAndSize ()
+		{
+			using (var win = new Window ()) {
+				win.InitialLocation = WindowLocation.Manual;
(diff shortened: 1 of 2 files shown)


sevoku:
Has it been solved by #393?

DavidKarlas:
Looks like it fixed but https://github.com/mono/xwt/blob/master/Xwt/Xwt/Window.cs#L251 is still wrong(shown = true is called few lines above).

sevoku:
True, but not really a bug (shown is always true there)(?)
Have you tried your unit tests against the current master? Maybe they could be merged.

DavidKarlas:
This 3 tests...
* Wpf: Works OK
* Gtk:
 * Windows: Unit tests are not running(BadImageFormat...)(Changing stuff to x86 didn't help)
 * Mac: Works OK
 * Linux: Works OK
* MonoMac: [ToDesktopRect](https://github.com/mono/xwt/blob/master/Xwt.Mac/Xwt.Mac/MacDesktopBackend.cs#L79) changes [Y from 230 to 548](https://github.com/DavidKarlas/xwt/blob/0f6c6b9f59950b57e096fbdf453790ff6604a4bf/Testing/Tests/WindowTests.cs#L243)(because I have MacBook with 900 pixel hight)

sevoku:
Ok, thanks! Could you create PRs (and split the commit) for the #L251 fix and maybe the unit tests? This one can be closed then.

Does the Mac ToDesktopRect need a fix (I have no Mac, testing my stuff in a VM)?

Looks like the GtkTestRunner needs a separate project configuration for windows.

DavidKarlas:
About ToDesktopRect... Not sure what is reason for that inverse calculating of Y...

PR #285, opened 2014-02-02
Fix for InitializeAsGuest for Mac

HarlanHugh:
These changes avoid the redundant call to NSApplication.Init() when running as guest, which causes the app to fail.

Changes to Application are needed because IsGuest does not get set properly because Application.InitializeAsGuest calls Initialize which calls Toolkit.Load with isGuest hardcoded to false.

### DIFF
--- a/Xwt.Mac/Xwt.Mac/MacEngine.cs
+++ b/Xwt.Mac/Xwt.Mac/MacEngine.cs
@@ -47,9 +47,9 @@ namespace Xwt.Mac
 		
 		public override void InitializeApplication ()
 		{
-      if(!IsGuest) {
+			if(!IsGuest) {
   			NSApplication.Init ();
-      }
+			}
 			//Hijack ();
 			if (pool != null)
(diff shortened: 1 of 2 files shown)


sevoku:
This looks correct to me. Could one of the Mac experts review this? Doesn't MD use Xwt in guest mode on Mac and why does it work there, if the initialization fails?

sevoku:
Update: I've looked into how MonoDevelop uses Xwt and it looks like its always using Gtk. The problem doesn't occur there, because `Gtk.Application.Init ()` can be called more then once. So we can use this but the default for isGuest should be false.

slluis:
This breaks MD. A similar patch was committed in the past and I had to revert. Using XWT.Mac as guest doesn't mean that Application.Init() has been called.

sevoku:
Ah, ok ha forgot to look into the MacPlatform addin. So what would be the correct way to initialize Xwt in this case?

PR #290, opened 2014-02-15
WebView improvements

bratsche:
This implements `NavigatingToUrl` and `NavigatedToUrl` events and `LoadHtmlString()` method.


sevoku:
I think we have a working solution with complete navigation features now (#341, #356).

PR #310, opened 2014-03-20
[GTK] ListView gets keyboard focus when SetFocus is called.

mrward:
Calling ListView.SetFocus() was not getting the keyboard focus so the up/down arrow keys did not select the next list view item.

### DIFF
--- a/Xwt.Gtk/Xwt.GtkBackend/ListViewBackend.cs
+++ b/Xwt.Gtk/Xwt.GtkBackend/ListViewBackend.cs
@@ -155,6 +155,12 @@ namespace Xwt.GtkBackend
 
 			return new Rectangle (x, y, w, h);
 		}
+
+		public override void SetFocus ()
+		{
+			base.SetFocus ();
+			Widget.GrabFocus ();
+		}
 	}
 }
 


sevoku:
According to the GTK Api (https://developer.gnome.org/gtk3/stable/GtkWidget.html#gtk-widget-is-focus) IsFocus is only to determine the focus state, so to really set the focus, GrabFocus should be called, if the Widget.CanFocus is true (and only then). So I think SetFocus can be uncommented in https://github.com/mono/xwt/blob/listview-focus/Xwt.Gtk/Xwt.GtkBackend/WidgetBackend.cs#L167 and a check for CanGetFocus should be added to SetFocus in https://github.com/mono/xwt/blob/listview-focus/Xwt.Gtk/Xwt.GtkBackend/WidgetBackend.cs#L174

sevoku:
@slluis: why is the parents SetFocus called here: https://github.com/mono/xwt/blob/listview-focus/Xwt.Gtk/Xwt.GtkBackend/WidgetBackend.cs#L174

sevoku:
I've added #473 that always calls GrabFocus. I think it should work for all widgets.

sevoku:
I've merged #473 (passes focus tests), so this can be closed.

Issue #314, opened 2014-04-01
[Xwt.Mac.Cocoa] - TreeView, ListView changing data in DataSource without effect


Strnadj:
On WPF and GTK it works fine! But when you're trying to add some items, this action has no effect when you're using MonoMac (Cocoa). You can actually see it in demo application for cocoa already, when you are trying Widgets -> TreeView (add some items / remove some items).

I need it, so I will try to fix it, but Iam not sure If am able to do it...

PR #316, opened 2014-04-01
Issue #314 - add reloading data source after source was changed, add exa...

Strnadj:
Reference issue #314


sevoku:
I close this in favor to #426

Strnadj:
Thanks :+1:

Issue #315, opened 2014-04-01
[Xwt.Mac.Cocoa] - Context Menu not working


Strnadj:
Can be see it in examples - Widgets - Menu.

sevoku:
fixed in #467

PR #319, opened 2014-04-03
#315 - Extends NSOutlineView for proper handling mouse events

Strnadj:
Iam referencing my Issue #315 from thursday, I have to fix this issue, cause I need it ;) It's something like "example solution", Iam new in Mono / Mono Mac / c#, but it works... I try to find mouse handler for List / TreeView but I was no able to find it... I have to extend NSOutlineView, like "Hywel Thomas" does it in "WidgetView".

### DIFF
--- a/TestApps/Samples/MainWindow.cs
+++ b/TestApps/Samples/MainWindow.cs
@@ -13,7 +13,9 @@ namespace Samples
 		VBox sampleBox;
 		Label title;
 		Widget currentSample;
-		
+		Menu contextMenu;
+		Label treeViewLabel;
+
 		DataField<string> nameCol = new DataField<string> ();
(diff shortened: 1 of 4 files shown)


sevoku:
Isn't this the same as #328 but for TreeViews? Could you move the sample from the main window to the TreeViews sample?

Strnadj:
@sevoku Yes its actually is, Iam now in middle of sth, I accidentally deleted my fork of mono/xwt. I will move examples to treeview, rebase it and create new PR, but probably tomorrow..

sevoku:
Sure, no problem! Take the time you need!

Strnadj:
Close this, accidentaly deleted XWT branch...

PR #340, opened 2014-06-05
Add notebook features (Icons, expanding tabs)

sevoku:
This one adds two useful features to the Notebook widget:
- Icons in Tab headers: makes it possible to add an icon in front of a tab label
- Expandable Tabs: if enabled, expands the tab lables

~~TODO: MAC backend support~~ (expanding doesn't really fit on MAC)
~~TODO: WPF expands the tabs only on initialization and resize.~~ fixed: 30c462765d4c49e376b2065e58bdb19b4bd9ea6f

### DIFF
--- a/Xwt.WPF/Xwt.WPFBackend/NotebookBackend.cs
+++ b/Xwt.WPF/Xwt.WPFBackend/NotebookBackend.cs
@@ -193,7 +193,17 @@ namespace Xwt.WPFBackend
 	{
 		public WidgetBackend Backend { get; set; }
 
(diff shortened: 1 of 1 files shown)


sevoku:
I was able to fix the immediate expansion change on Wpf. So everything works except expanding on Mac, which is ok, since the tabs are very wide on Mac by design and expanding them doesn't really fit.

Thats how it looks like (expanded with icons):
xwt demo application_041
xwp wpf expand notebook tabs
xwp mac notebook tabs icons

PR #363, opened 2014-07-17
Add multithreading patterns

sevoku:
I've added some multithreading features including a "Multithreading" sample, which starts different worker patterns and invokes results in the gui thread.

**XwtSynchronizationContext**: is installed automatically and uses Application.Invoke for synchronization. BackgroundWorker can synchronize with the UI Thread automatically for example.

**XwtComponent**: implements the old fashioned ISynchronizeInvoke. Very useful when porting old WinForms Implementations to Xwt.

**Possible regressions**: should be tested in Guest mode

I've tested everything on Linux, Windows (Wpf and Gtk) and Mac (Cocoa and Gtk)

PS: let me know if I should split this in different PRs. The first commit simply adds the sample part without the new implementations and could be very useful for others.

### DIFF
--- a/Xwt/Xwt/XwtComponent.cs
+++ b/Xwt/Xwt/XwtComponent.cs
@@ -80,14 +80,14 @@ namespace Xwt
 
 		#region ISynchronizeInvoke implementation
 
-		public IAsyncResult BeginInvoke (Delegate method, object[] args)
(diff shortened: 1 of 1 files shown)


sevoku:
Strange, Github shows "Closed with unmerged commits", but everything seems to be merged. Github bug?

PR #367, opened 2014-07-23
[Gtk] TextEntryBackend with MultiLine-Feature

lytico:
TextEntryMultiLineBackend: a backent that implements Xwt.TextEntry.MultiLine

It is based on Gtk.TextView instead of Gtk.TextEntry


lytico:
I think this can be closed in favor to 
https://github.com/mono/xwt/pull/375

PR #370, opened 2014-07-29
[WPF] Make sure that the dialog appears on top, instead of behind other windows

steffenWi:
A dialog should appear above other windows when opened.

If this is not desired, please leave a comment on how to achieve a similiar result since 'TopMost' is not exposed/part to/of the XWT API.


sevoku:
I've added a new feature request (#454). This one can be closed.

PR #377, opened 2014-08-13
[Gtk] fix cursor handling

sevoku:
The cursor should be set only when inside the widget.
Set the cursor when mouse enters and reset it to parents
default (null in Gtk) when mouse leaves the widget.

(fix can be reproduced on the Cursors example with Gtk2)

### DIFF
--- a/Xwt.Gtk/Xwt.GtkBackend/WidgetBackend.cs
+++ b/Xwt.Gtk/Xwt.GtkBackend/WidgetBackend.cs
@@ -225,25 +225,29 @@ namespace Xwt.GtkBackend
 
 			gdkCursor = gc;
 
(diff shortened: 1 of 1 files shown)


sevoku:
Hhm, maybe this should only be made on Gtk2. Gtk3 works correctly. I'll review this again, and report.

sevoku:
Ok, the cursors example works with Gtk3 without this change (the labels have own windows in Gtk3, in Gtk2 not). But there can be still situations when a widget has no own GdkWindow even with Gtk3. So this PR can be merged.

sevoku:
I've tested this now additionally on Win and Mac. Everything works as expected.

PR #381, opened 2014-08-14
[Gtk] fix evt handling for widgets with own window

sevoku:
Some Gtk Widgets have their own Gdk.Window. Most of the events are not raised for such widgets, because WidgetBackend (and its derivates) automatically put them into an EventBox and subscribe its events, instead of using the widgets directly.

This PR allows Gtk backends to overide EventBox allocation and handle events directly for widgets which have a Gdk.Window already. This is achieved though the WidgetBackend.NeedsEventBox property, which is true by default (keeps original behavior). Backends can set it to false during their initialization and force the base class WidgetBackend to subscribe the events from the widgets directly.

I've tested almost all widgets (but not all events, only Mouse events) and added the override to widgets, that did not work previously. Maybe some other backends should use the override, too.

### DIFF
--- a/Xwt.Gtk/Xwt.Gtk3.csproj
+++ b/Xwt.Gtk/Xwt.Gtk3.csproj
@@ -150,6 +150,7 @@
     <Compile Include="Xwt.GtkBackend\GtkWebKitMini.cs" />
     <Compile Include="Xwt.GtkBackend\PopoverBackend.cs" />
     <Compile Include="Xwt.GtkBackend\Gtk3PopoverWindow.cs" />
+    <Compile Include="Xwt.GtkBackend\PanedBackendGtk3.cs" />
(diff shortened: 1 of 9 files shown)


sevoku:
I've found the no-eventbox branch (b42c496) where @slluis started some initial work to eliminate EventBox usage completely, which would be a much better solution. But it looks like there is still much work to be done. Until then this PR is a good temp solution.

slluis:
There isn't in fact so much work left to be done, but it requires a lot of testing.

sevoku:
Hhm, I had not much time to take a closer look and test it. How should we proceed? I'd propose to merge this for a working (temporally) solution making things work and to finish your work on eventbox elimination for 1.0 (or maybe earlier).

slluis:
I like the plan.

PR #384, opened 2014-08-22
TextInput event optimization

sevoku:
Consistency: rename PreviewTextInput event to TextInput
* Xwt has no preview events. To stop an event, the Handled property of the event arguments object should be used.

GTK: adds TextInput event support
* different implementation for Editable and non-Editable widgets
* Gtk.Editable (like Entry, TextView, etc.): simply handle TextInserted event
* Other widgets: use Gtk.IMMulticontext to get text input directly from the system IM

### DIFF
--- a/TestApps/Samples/Samples/TextInput.cs
+++ b/TestApps/Samples/Samples/TextInput.cs
@@ -15,7 +15,7 @@ namespace Samples
 
 			public TextEditor ()
 			{
-				this.PreviewTextInput += HandleTextInput;
+				this.TextInput += HandleTextInput;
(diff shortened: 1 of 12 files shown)


cra0zy:
@sevoku Will this get merged any time soon? I am fixing up XwtPlus TextEditor and this is the only thing that is needed.

sevoku:
Hhm, this needs a rebase...

I didn't merge this cause it changes the API and I didn't want to do that without feedback from @slluis. I'll rebase and merge next days.

cra0zy:
Thanks for the info and quick response.

PR #388, opened 2014-09-02
Add SpinButton custom formatting support + backend fixes

sevoku:
Adds ValueInput, ValueOutput events for custom value-text, text-value conversion.
Additionally the shown text can be manipulated through the new Text property.

This makes the SpinButton much more flexible. The included examples show spin buttons with different number formats like hex or scientific floats. The new events can be used to convert the numeric value to any text output and to parse the users input back to a numeric value. The input text can be accessed directly for this purpose.

Xwt SpinButton Demo

The pattern is similar to Gtk. A change of the numeric value triggers the ValueOutput event. The handler must do the conversion, assign the Text property and set args.Handled. Otherwise the conversion is handled by the backend. A user input (text change) triggers ValueInput. The handler must assign the args.NewValue property of the event args object and set args.Handled. 

If the events are not handled (args.Handled = false, or nor even subscribed), the backends behave as usual.

### DIFF
--- a/Xwt.WPF/Xwt.WPFBackend/WindowsSpinButton.xaml.cs
+++ b/Xwt.WPF/Xwt.WPFBackend/WindowsSpinButton.xaml.cs
@@ -89,10 +89,21 @@ namespace Xwt.WPFBackend
         {
             get { return (double)GetValue(ValueProperty); }
             set {
-                SetValue(ValueProperty, Math.Round(value, DecimalPlaces));
+				SetValue(ValueProperty, SnapToTick(value));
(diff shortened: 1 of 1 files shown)


sevoku:
I've rebased this against the current master with the recent XamMac merges. This PR includes some important MAC fixes, which do not depend on the custom formatting support itself. Should I create a separate PR for them (needs some work to split them, since I applied the after adding custom formatting) so it can be reviewed separately? Or can we merge this as it is?

sevoku:
I've extracted the Mac fixes and added some more to #505.

PR #389, opened 2014-09-05
[Gtk] fix pointer coordinates for mouse events

sevoku:
**EDIT:** Some widgets have additional child Gdk windows (or real child widgets if the widget is a container) on top of their root window. In this case pointer events may come from child windows and contain wrong coordinates (relative to child window and not to the widget itself). 
The CheckPointerCoordinates workaround function checks whether the events source window is not the widgets root window and adjusts the coordinates to be relative to the widget and not to its child.

*Old message: If a widget is a container and a mouse event is raised by one of its children with own child Gdk.Window, the pointer coordinates are relative to the child and not the widget itself. In this case the pointer coordinates should be updated.*

### DIFF
--- a/Xwt.Gtk/Xwt.GtkBackend/GtkWorkarounds.cs
+++ b/Xwt.Gtk/Xwt.GtkBackend/GtkWorkarounds.cs
@@ -1226,6 +1226,36 @@ namespace Xwt.GtkBackend
 		}
 
 
+		/// <summary>
+		/// Adjusts pointer coordinates to be relative to the widget.
+		/// </summary>
+		/// <returns>The pointer coordinates.</returns>
+		/// <param name="widget">The Widget.</param>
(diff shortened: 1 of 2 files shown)


sevoku:
I've rebased the branch and removed the container condition. There are other widgets affected by this bug, like e.g. SpinButton or Paned. Additionally I've added some documentation to the code.

sevoku:
The problem can be reproduced by adding
```c#
this.MouseMoved += (sender, e) => Console.WriteLine (e.X + "x" + e.Y);
```
to the Samples.WidgetEvents constructor (https://github.com/mono/xwt/blob/master/TestApps/Samples/Samples/WidgetEvents.cs#L74) and moving the mouse over the events examle vbox.
Without the patch the Y coordinate is wrong, when the mouse is moved over the widgets inside the box.

PR #391, opened 2014-09-07
[Gtk] Optimize message dialogs

sevoku:
Let Gtk.MessageDialog do the layout stuff and make use of Gtk.MessageType for a more native icon styles.

This change makes message dialogs look more native and closer to gtk design guidelines (especially with gtk3). Additionally the dialogs should be compatible with more themes (providing special designs for message dialogs).

(an extended message dialog sample has been added to the sample app, too: #380)

Some example screenshots (Ubuntu - Gnome 3):

Before (gtk2/3)| After (gtk2) | After (gtk3)
------------ | ------------- | --------------
error_old | error_new_gtk2 | error_gtk3
question_old | question_gtk2 | question_gtk3


sevoku:
I've rebased this PR and removed the icon handling fix from #392.

slluis:
Looks good to me.

PR #392, opened 2014-09-08
[Gtk] Fix default icon sizes

sevoku:
ImageDescription.WithDefaultSize should always return the icon with the default size.
Backends like Button or ImageRenderer should decide to enforce a default size, or to set it only, if not done by the user.

(78f999f follow up)


sevoku:
Not really required with the new MessageDialog (#391).

PR #409, opened 2014-12-09
[WPF] fix canvas mouse events

sevoku:
Initialize the background of the underlying Panel with a transparent background.
A Panel does not receive mouse events, if no background is set.

(see MSDN remarks: http://msdn.microsoft.com/en-us/library/system.windows.controls.panel(v=vs.110).aspx)


slluis:
I feel this should be set only when mouse events are subscribed. Other than that, it looks good.

sevoku:
Hhm, then the background would be reseted on every subscription and break the Background property. Setting it once at initialization makes it possible to change the background afterwards as desired and does not break anything (needing a much more complex logic for event vs background handling).

PR #422, opened 2014-12-20
Row height for TreeView and ListView

residuum:
- Using Try...(out ...) idiom
- Samples included
- WPF implementation
- GTK implementation
- GTK3 implementation
- Mac returns false


sevoku:
Hi! Could you try #433? It contains the complete implementation discussed earlier. You could use GetRowBounds (on all Tree/List/ListBox), which gives you the bounds relative to the widget itself. If the widget has scroll controls, you will have to adjust the bound position. An example with a dynamic tracker (works only with Gtk3, since it needs canvas alpha) is included, to get an idea of how to use it.

residuum:
Yes, your implementation superseeds this PR, great job.

PR #423, opened 2014-12-20
Refine DatePicker API and impement DatePickerStyle support [Gtk][Mac]

sevoku:
This PR adds DatePickerStyle support to Gtk and Mac backends, and includes many fixes in the backend implementations.

* GTK:
  * fix component selection
  * fix (keep) seconds when other component chages
  * fix keyboard input (incl. arrow keys and separators)
  * fix mouse selection
  * add internationalization support
    * use localized separators
    * use localized component order
    * no 12/24 support
  * add DatePickerStyle support (show only relevant components)
  * add clipboard support
* MAC:
  * add DatePickerStyle support
  * fix DatePicker timezone conversion

For WPF we still need a useful DateTimePicker component. Maybe something external from NuGet?

### DIFF
--- a/Xwt.Gtk/Xwt.GtkBackend/DatePickerBackend.cs
+++ b/Xwt.Gtk/Xwt.GtkBackend/DatePickerBackend.cs
@@ -60,21 +60,21 @@ namespace Xwt.GtkBackend
 			}
 		}
 
-		public DateTime MinDateTime {
+		public DateTime MinimumDateTime {
 			get {
-				return Widget.MinDateTime;
+				return Widget.MinimumDateTime;
 			}
 			set {
-				Widget.MinDateTime = value;
+				Widget.MinimumDateTime = value;
 			}
 		}
 
-		public DateTime MaxDateTime {
+		public DateTime MaximumDateTime {
 			get {
-				return Widget.MaxDateTime;
+				return Widget.MaximumDateTime;
 			}
 			set {
-				Widget.MaxDateTime = value;
+				Widget.MaximumDateTime = value;
 			}
 		}
 
@@ -148,7 +148,7 @@ namespace Xwt.GtkBackend
 				}
 			}
 
-			public DateTime MinDateTime {
+			public DateTime MinimumDateTime {
 				get {
 					return new DateTime ((long)Adjustment.Lower);
 				}
@@ -159,7 +159,7 @@ namespace Xwt.GtkBackend
 				}
 			}
 
-			public DateTime MaxDateTime {
+			public DateTime MaximumDateTime {
 				get {
 					return new DateTime ((long)Adjustment.Upper);
 				}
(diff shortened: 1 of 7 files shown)


Xwt.Gtk/Xwt.GtkBackend/DatePickerBackend.cs
```diff
@@ -85,253 +96,348 @@ void HandleValueChanged (object sender, EventArgs e)
...
+					return style;
+				}
+				set {
+					style = value;
+
+					string format = CultureInfo.CurrentCulture.DateTimeFormat.ShortDatePattern +
+						" " + "HH" + timeSeparator + "mm" + timeSeparator + "ss";
+
+					if (styleFormats.ContainsKey (value))
+						format = styleFormats [value];
```
slluis: TryGetValue should be used here

Xwt.Gtk/Xwt.GtkBackend/DatePickerBackend.cs
```diff
@@ -85,253 +96,348 @@ void HandleValueChanged (object sender, EventArgs e)
...
-			int currentDigitInsert;
-	
-			public GtkDatePickerEntry () : base (DateTime.MinValue.Ticks,
-			                                     DateTime.MaxValue.Ticks,
-			                                     TimeSpan.TicksPerSecond)
+			public GtkDatePickerEntry (DatePickerStyle style) : base (DateTime.MinValue.Ticks,
+			                                                          DateTime.MaxValue.Ticks,
+			                                                          TimeSpan.TicksPerSecond)
 			{
+				styleFormats[DatePickerStyle.Date] = CultureInfo.CurrentCulture.DateTimeFormat.ShortDatePattern;
```
slluis: Shouldn't styleFormats be static?

Xwt.Gtk/Xwt.GtkBackend/DatePickerBackend.cs
```diff
@@ -85,253 +96,348 @@ void HandleValueChanged (object sender, EventArgs e)
...
 			protected override int OnOutput ()
 			{
-				DateTime dateTime = CurrentValue;
-				Text = dateTime.ToString (DateTimeFormat);
-				
+				DateTime dateTime = new DateTime ((long)Adjustment.Value);
+				if (styleFormats.ContainsKey (Style))
+					Text = dateTime.ToString (styleFormats [Style]);
+				else
+					Text = CultureInfo.CurrentCulture.DateTimeFormat.ShortDatePattern
```
slluis: We don't need to support this scenario. The style should be a valid enum value.

sevoku:
You're right about the formats in Gtk. I've made them static and did some cleanup.

sevoku:
I've added an other timezone fix on Mac, removed the static dictionary from the Gtk backend and renamed the Min/Max values (consistency). Its ok to merge now.

PR #426, opened 2014-12-29
[Mac] reload tree/list on data source changes

### LINKED ISSUE #314: [Xwt.Mac.Cocoa] - TreeView, ListView changing data in DataSource without effect

sevoku:
This is the same as ##316, but reloads the views not only on row changes, but also on  deletion/insertion.

(closes #314, closes #316)

### DIFF
--- a/TestApps/Samples/Samples/ListView1.cs
+++ b/TestApps/Samples/Samples/ListView1.cs
@@ -52,6 +52,27 @@ namespace Samples
 				}
 			};
 
+			HBox btnBox = new HBox ();
+			Button btnAddItem = new Button ("Add item");
+			btnAddItem.Clicked += delegate {
+				var r = store.InsertRowAfter(list.SelectedRow < 0 ? 0 : list.SelectedRow);
+				store.SetValue (r, icon, png);
+				store.SetValue (r, name, "Value " + (store.RowCount + 1));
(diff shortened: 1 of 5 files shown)


sevoku:
I've tried to optimize the performance of source operations.
Now only the affected nodes are reloaded in trees, which is a big win, since reloading the complete tree need all TreeItem instances to be recreated. As a side effect the row expansions and selections are kept intact.
Lists optimization is a little bit more difficult for RowChanged, but all other store changes seem to always need a data reload (according to [NSTableView docs](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/ApplicationKit/Classes/NSTableView_Class/#//apple_ref/occ/instm/NSTableView/reloadData) the NSTableView.ReloadData() reloads only the visible items and there are no view wrappers like in TreeViews, so it is not really a big impact on performance).

@Strnadj: could you take a look and try whether it works as expected (compared to #316)?

Strnadj:
Sorry, I didn´t noticed you asked me, I will try it probably tomorrow.

PR #429, opened 2014-12-29
Gtk tree ITreeDataSource events, cleanup and various fixes

sevoku:
This PR includes a ITreeDataSource events implementation for TreeStoreBackend and some code cleanup.

PR #430, opened 2014-12-29
Add tree filtering through proxy stores

sevoku:
This PR adds new ListStoreFilter and TreeStoreFilter stores, which can be used to filter other stores (IListDataSource, ITreeDataSource). For this purpose they wrap around an other store and pass a filtered version to a view (similar to Gtk implementation). A filter function can be specified, which will be called for every node/row to evaluate whether to filter (hide) it or not.

Exaples for a filtered ListView and TreeView are included.

Backends:
Gtk: native implementation using Gtk.TreeModelFilter
Wpf: ~~works out of the box~~ works partially (needs review)
Mac: uses default Xwt backends (both included), needs #426 (!)

Other changes:
* depends on / includes #428, (partially) #429
* Xwt: ITreeDataSource / IListDataSource implementations moved to Xwt.*StoreBase classes
* Gtk: ITreeDataSource / IListDataSource implementations moved to *StoreBase classes

### DIFF
--- a/TestApps/Samples/MainWindow.cs
+++ b/TestApps/Samples/MainWindow.cs
@@ -83,6 +83,7 @@ namespace Samples
 			AddSample (w, "ListBox", typeof(ListBoxSample));
 			AddSample (w, "LinkLabels", typeof(LinkLabels));
 			var listView = AddSample (w, "ListView", typeof(ListView1));
+			AddSample (listView, "Filter", typeof(ListViewFilter));
 			AddSample (listView, "Editable checkboxes", typeof(ListView2));
 			AddSample (w, "Markdown", typeof (MarkDownSample));
 			AddSample (w, "Menu", typeof(MenuSamples));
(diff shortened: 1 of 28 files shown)


slluis:
Why not just add filtering capabilities to the existing stores?

sevoku:
I was thinking about that, but had some problems with the implementations:
1. different DataStore interfaces are required (one used directly by the frontend, and the other one passed to the tree backend)
2. Gtk: the logic has to be much more complex, since we would have a backend (the TreeModelFilter) and a frontend (the unfiltered model) side. Adding/removing/changing nodes is not an issue, but Selections etc. need some sort of wrapping (by implementing different interfaces, 1.).
4. Mapping row ids in ListViews needs even more work (need separate tracking)
3. The default implementation was too much work for me

At the other hand I really like the Gtk pattern with wrapping the Models. The user does not need to think about which node he is working with (filtered/unfiltered) when accessing the tree widget itself. He must only use the conversion functions.

I think both approaches can live together. We could add filtering to the Stores directly some day and it would not interfere with the proxy approach.

sevoku:
WPF backend has some issues. An INotifyCollectionChanged wrapper  is needed for TreeStore.

sevoku:
Hhm, there seem to some more problems with the Wpf backend and big performance issues on Mac and Wpf (generally for all trees on mac and for foreign data sources on Wpf). It has nothing to do with this PR directly, but it depends on store events implementation, which is not sophisticated enough.
I'll create a new PR (with optimized store events, #426, and store fixes for Wpf), and rebase this one to match the changes. Until then I close this one (will reopen, when ready).

PR #431, opened 2015-01-03
[Gtk] context fix for Gtk2 and Gtk3

hwthomas:
Gtk3 DrawingTransforms sample does not draw correctly with Gtk3 at present.  This is because the context origin has to be adjusted, so that the new Context.CTM is always the Identity Matrix.  In addition, the invalidated area supplied to the OnDraw routines needs adjusting for the same reason.

### DIFF
--- a/Xwt.Gtk/Xwt.CairoBackend/CairoContextBackendHandler.cs
+++ b/Xwt.Gtk/Xwt.CairoBackend/CairoContextBackendHandler.cs
@@ -368,12 +368,9 @@ namespace Xwt.CairoBackend
 		public override Matrix GetCTM (object backend)
 		{
 			var cb = (CairoContextBackend)backend;
-
 			Cairo.Matrix t = cb.Context.Matrix;
-
-			// Adjust CTM OffsetX, OffsetY for ContextBackend Origin
+			// Adjust CTM X0,Y0 for ContextBackend Origin (ensures that new CTM is Identity Matrix)
 			Matrix ctm = new Matrix (t.Xx, t.Yx, t.Xy, t.Yy, t.X0-cb.Origin.X, t.Y0-cb.Origin.Y);
-
 			return ctm;
 		}
 
--- a/Xwt.Gtk/Xwt.GtkBackend.CellViews/Gtk2CellRendererCustom.cs
+++ b/Xwt.Gtk/Xwt.GtkBackend.CellViews/Gtk2CellRendererCustom.cs
@@ -31,8 +31,11 @@ namespace Xwt.GtkBackend
 	{
 		protected override void Render (Gdk.Drawable window, Gtk.Widget widget, Gdk.Rectangle background_area, Gdk.Rectangle cell_area, Gdk.Rectangle expose_area, CellRendererState flags)
 		{
-			using (var cr = Gdk.CairoHelper.Create (window))
+			using (var cr = Gdk.CairoHelper.Create (window)) {
+				cr.Rectangle (background_area.X, background_area.Y, background_area.Width, background_area.Height);
+				cr.Clip ();
 				OnRender (cr, widget, background_area, cell_area, flags);
+			}
 		}
 
 		protected abstract void OnRender (Cairo.Context cr, Gtk.Widget widget, Gdk.Rectangle background_area, Gdk.Rectangle cell_area, CellRendererState flags);
--- a/Xwt.Gtk/Xwt.GtkBackend/CanvasBackend.cs
+++ b/Xwt.Gtk/Xwt.GtkBackend/CanvasBackend.cs
@@ -160,8 +160,7 @@ namespace Xwt.GtkBackend
 		{
 			Backend.ApplicationContext.InvokeUserCode (delegate {
 				using (context) {
-					var r = new Rectangle (dirtyRect.X - context.Origin.X, dirtyRect.Y - context.Origin.Y, dirtyRect.Width, dirtyRect.Height);
-					EventSink.OnDraw (context, r);
+					EventSink.OnDraw (context, dirtyRect);
 				}
 			});
 		}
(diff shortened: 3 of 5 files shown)


hwthomas:
This problem is identical (but fot Gtk3) to those fixed by PR's 337 and 272.  Gtk3, unlike Gtk2, does not provide access to the invalidated area when a QueueDraw (rect) is done, so the full Canvas Allocation has to be used for drawing.  The Clip rect *is*, however, set correctly (internally), so any drawing is limited to the specified rect.  I've had to rearrange the code between CanvasBackend, CanvasBackendGtk2, and CanvasBackendGtk3 in order to do the offsets only where necessary.  The issues would have been picked up by the TransformTests, but I notice that there is no Gtk3TestRunner yet (any reason? I'll certainly help if necessary).  I've tested fairly carefully, but an extra drawing sample might be useful (if anyone else knows what to look for).

sevoku:
Hhm, yes! I didn't realize that before.
Before:
xwt demo application_034
After:
xwt demo application_035

sevoku:
Something different, but I asked myself why the canvas draws sometimes outside its bounds with Gtk2, but not with Gtk3. As you have commented, Gtk3 clips the context and Gtk2 doesn't. Maybe you could add clipping to Gtk2, too?

hwthomas:
Do you have any examples where the clipping *doesn't* work?  I did some quick tests on this and both Gtk2 and Gtk3 seem to enforce the clip limits set by QueueDraw (dirtyRect), even though only Gtk2 passes this rect to the OnDrawn routine.  A quick-and-dirty test (replacing the DrawingTransforms code) is:-
```
public class DrawingTransforms: Canvas
{
	//Test code to check clipping
	private bool MouseDown = false;

	protected override void OnDraw (Context ctx, Rectangle dirtyRect)
	{
		ctx.Rectangle (0.5, 0.5, 43, 33);	// draw blue rectangle outline
		ctx.SetColor (Colors.Blue);
		ctx.SetLineWidth (1);
		ctx.Stroke ();
		if (MouseDown) {
			ctx.Rectangle (-10, -10, 200, 200);	// deliberately exceed dirtyRect
			ctx.SetColor (Colors.Red);
			ctx.Fill ();
		}
	}

	protected override void OnButtonPressed (ButtonEventArgs args)
	{
		MouseDown = true;
		// Invalidate an area of canvas 1 pixel within blue rectangle when clicked
		Rectangle r = new Rectangle (2, 2, 40, 30);
		QueueDraw (r);	// then redraw just this area
	}

	protected override void OnButtonReleased (ButtonEventArgs args)
	{
		MouseDown = false;
	}
	//End Test code
}
```
This works OK for me.

sevoku:
You can see it in the "Scroll View" sample when you resize the window or in my new Cell Bounds sample (#433), but only with Gtk2.
```c#
protected void OnDraw (Rectangle dirtyRect, CairoContextBackend context)
{
	Backend.ApplicationContext.InvokeUserCode (delegate {
		using (context) {
			var r = new Rectangle (dirtyRect.X - context.Origin.X, dirtyRect.Y - context.Origin.Y, dirtyRect.Width, dirtyRect.Height);
+			context.Context.Rectangle (dirtyRect.X - context.Origin.X, dirtyRect.Y - context.Origin.Y, dirtyRect.Width, dirtyRect.Height);
+			context.Context.Clip();
			EventSink.OnDraw (context, r);
		}
	});
}
```
Maybe a `context.Context.ResetClip()` should come after `EventSink.OnDraw (context, r)`, but Gtk2 crashes then.

Some Screenshots:

Without Clipping:
xwt demo application_036

With Clipping:
xwt demo application_037

sevoku:
Hhm, it breaks Gtk3. The clipping should go into CanvasBackendGtk2.cs

hwthomas:
Odd.  I get different results.  I can't get Gtk2 to draw outside the canvas Bounds (just by hacking my example above).  The ScrollView sample *does* draw over the scroll bar area (which looks more like a ScrollableCanvas issue), but *doesn't* extend as far as your screenshot above.  In Gtk3, I don't even get OnDrawn in the CanvasBackendGtk3 called! (so OnDraw in the ScrollableCanvas is never called). I'm using Monodevelop  5.0.1, and the latest sources, but at the moment I have no idea what is happening.

sevoku:
Hhm, thats strange. Drawing over ScrollBar should not happen. The only ScrollBar isuue is the little tiny rectangle in the bottom right corner (but thats normal for Gtk, since the scroll controls are overlays above the child). Maybe some older Gtk/Cairo issue fixed in newer versions (you're using an older MD version)?

I've tested on Ubuntu 14.10 (Gnome 3.14) with MD 5.5 (complete mono from Xamarin http://origin-download.mono-project.com/repo/debian/) and Window 7 with XS 5.5 and GTK 3.6.4.


the definitive clipping fix should be in CanvasBackendGtk2.cs:
```c#
protected override bool OnExposeEvent (Gdk.EventExpose evnt)
{
	var a = evnt.Area;
	var dirtyRect = new Rectangle (a.X, a.Y, a.Width, a.Height);
	using (var ctx = CreateContext ()) {
		ctx.Context.Rectangle (dirtyRect.X - ctx.Origin.X, dirtyRect.Y - ctx.Origin.Y, dirtyRect.Width, dirtyRect.Height);
		ctx.Context.Clip ();
		OnDraw (new Rectangle (a.X, a.Y, a.Width, a.Height), ctx);
	}
	return base.OnExposeEvent (evnt);
}
```

additionally to your changes.

hwthomas:
I agree that the clipping can be done as above, but the ExposeEvent area needs the ctx.Origin offsetting before passing to OnDraw, ie:-
```
protected override bool OnExposeEvent (Gdk.EventExpose evnt)
{
    var a = evnt.Area;
    var dirtyRect = new Rectangle ((a.X - ctx.Origin.X, a.Y - ctx.Origin.Y, a.Width, a.Height);
    using (var ctx = CreateContext ()) {
        ctx.Context.Rectangle (dirtyRect);
        ctx.Context.Clip ();
        OnDraw (dirtyRect, ctx);
    }
    return base.OnExposeEvent (evnt);
}
```
Do you agree?

sevoku:
Sure, you've done it in OnDraw (CanvasBackend.cs) already, which is called from the Gtk2/3 base classes. Gtk2 base clips additionally using the offset with my patch. So for Gtk3 it is already done in its derivate (your patch).

I could add a PR against your repo so you could merge it into this PR, if you like.

sevoku:
Sorry, forget my last comment! I checked out the wrong branch!

hwthomas:
Sorry, I was getting confused as to whether you were adding to my patches.  I've added another commit for the Gtk2 clipping which should, taken altogether, fix the issue

sevoku:
yes, it works! Thanks!

hwthomas:
This now adds the Gtk2 clipping changes from sevoku so that all changes are under the same topic. Thanks for all your comments and help, Vsevolod.

slluis:
Looks good to me

sevoku:
Ok, i'll merge this. Regarding the Gtk3 drawing problem described by @hwthomas ([#issuecomment-68861931](#issuecomment-68861931)), i've created a new issue #468. But it has nothing to do with this PR.

sevoku:
And: Thanks :+1:

PR #433, opened 2015-01-04
Extend Tree/List cell bounds API, various Gtk fixes, Mac impl.

sevoku:
This PR adds the already known ListView cell bounds retrieval functions to ListBox and TreeView. Additionally complete row (not only cell) bounds can be retrieved. ListView/TreeView examples included.

Backend implementations/support:
* Gtk: completely rewritten
  * now with real Gtk3 support
  * more native position calculation without static paddings
  * real data is loaded for calculation
  * includes #432 for better ```GetRowAtPosition``` support
* Mac: bounds always include margins (background bounds). Everything else works as expected (tested in VM, needs more testing on a real machine).
* Wpf: ~~still TODO!~~ done in c5b4130
  * static implementation by loading fixed FrameworkElement children
  * the view layout hierarchy may not change

It could be a more consequent solution for #422.

### DIFF
--- a/Xwt.XamMac/Xwt.Mac/TreeViewBackend.cs
+++ b/Xwt.XamMac/Xwt.Mac/TreeViewBackend.cs
@@ -32,6 +32,7 @@ using System.Collections.Generic;
 using MonoMac.AppKit;
 using MonoMac.Foundation;
 using nint = System.Int32;
+using nfloat = System.Single;
 using CGPoint = System.Drawing.PointF;
 #else
 using AppKit;
@@ -208,7 +209,7 @@ namespace Xwt.Mac
 
 		public TreePosition GetRowAtPosition (Point p)
 		{
-			var row = Table.GetRow (new System.Drawing.PointF ((float)p.X, (float)p.Y));
+			var row = Table.GetRow (new CGPoint ((float)p.X, (float)p.Y));
 			return row >= 0 ? ((TreeItem)Tree.ItemAtRow (row)).Position : null;
 		}
 
--- a/Xwt.XamMac/Xwt.Mac/TreeViewBackend.cs
+++ b/Xwt.XamMac/Xwt.Mac/TreeViewBackend.cs
@@ -208,17 +208,26 @@ namespace Xwt.Mac
 
 		public TreePosition GetRowAtPosition (Point p)
 		{
-			throw new NotImplementedException ();
+			var row = Table.GetRow (new System.Drawing.PointF ((float)p.X, (float)p.Y));
+			return ((TreeItem)Tree.ItemAtRow (row)).Position;
 		}
 
 		public Rectangle GetCellBounds (TreePosition pos, CellView cell, bool includeMargin)
 		{
-			throw new NotImplementedException ();
+			var it = tsource.GetItem (pos);
+			if (it == null)
+				return Rectangle.Zero;
+			var row = Tree.RowForItem (it);
+			return GetCellBounds (row, cell, includeMargin);
 		}
 
 		public Rectangle GetRowBounds (TreePosition pos, bool includeMargin)
 		{
-			throw new NotImplementedException ();
+			var it = tsource.GetItem (pos);
+			if (it == null)
+				return Rectangle.Zero;
+			var row = Tree.RowForItem (it);
+			return GetRowBounds (row, includeMargin);
 		}
 		
 		public bool GetDropTargetRow (double x, double y, out RowDropPosition pos, out TreePosition nodePosition)
--- a/Xwt.XamMac/Xwt.Mac/TreeViewBackend.cs
+++ b/Xwt.XamMac/Xwt.Mac/TreeViewBackend.cs
@@ -209,7 +209,7 @@ namespace Xwt.Mac
 		public TreePosition GetRowAtPosition (Point p)
 		{
 			var row = Table.GetRow (new System.Drawing.PointF ((float)p.X, (float)p.Y));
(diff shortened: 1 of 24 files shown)


sevoku:
Some screenshots of the included sample:
xwt demo application_030
xwt demo application_031

sevoku:
And some Wpf screenshots:
xwt demo application wpf listview bounds
xwt demo application wpf treeview bounds

residuum:
Wow, that looks great. It is working for me in Gtk2 and Gtk3, will test WPF tomorrow.

residuum:
Working with WPF as well.

Xwt.Mac/Xwt.Mac/TableViewBackend.cs
```diff
@@ -41,6 +42,13 @@ public abstract class TableViewBackend<T,S>: ViewBackend<NSScrollView,S>, ITable
 		ScrollView scroll;
 		NSObject selChangeObserver;
 		NormalClipView clipView;
+
+		Dictionary<CellView,CellInfo> cellViews = new Dictionary<CellView, CellInfo> ();
+
+		class CellInfo {
+			public ICellRenderer Cell;
+			public int Column;
+		}
```
slluis: I don't think we need a new class for this. This information can be stored in CellViewBackend.

sevoku: Hhm, maybe... I used the pattern from the Gtk implementation.

residuum:
I am currently working on fixing the issues with WPF on a seperate branch: https://github.com/residuum/xwt/tree/treeview-wpf-fixes

Xwt.Mac/Xwt.Mac/TableViewBackend.cs
sevoku: I've removed the Mapping in ec2daf9a2943fa62a162c703ada4238a759f6646. Now the column is retrieved directly from the CellViewBackend.

sevoku:
Rebased for XamMac, but it needs maybe a little review with Xamarin.Mac, since I can test it only with MonoMac.

And here is a Mac screenshot:
List cell bounds test on Mac

slluis:
I'm not seeing the correct bounds on Mac:

sevoku:
Hhm, that looks strange. There seems to be an issue with the bitmap generation of the selected row. Looks like the vertical row position is wrong. Does it always look like that even after scrolling up/down? Is it MonoMac or XamMac?

But the actual bounds look right. On Mac the overlay is not supported, so you should select a row at the end of the scroll window to (visually) compare the row with the drawn bounds.

An as I have written, I was not able to calculate the margins (see my PR description). So the cell bounds include only the rendered cell width and row height. You can see it if you switch between a line with a linebreak and one without (first column), the cell bound should change horizontally.

Xwt.XamMac/Xwt.Mac/TreeViewBackend.cs
```diff
@@ -192,6 +192,30 @@ public void ExpandToRow (TreePosition pos)
...
+			var row = Table.GetRow (new System.Drawing.PointF ((float)p.X, (float)p.Y));
+			return row >= 0 ? ((TreeItem)Tree.ItemAtRow (row)).Position : null;
+		}
+
+		public Rectangle GetCellBounds (TreePosition pos, CellView cell, bool includeMargin)
+		{
+			var it = tsource.GetItem (pos);
+			if (it == null)
+				return Rectangle.Zero;
+			var row = Tree.RowForItem (it);
```
slluis: The result of Tree.RowForItem has to be casted to int, since it returns an nint on XamMac.

Xwt.XamMac/Xwt.Mac/TreeViewBackend.cs
```diff
@@ -192,6 +192,30 @@ public void ExpandToRow (TreePosition pos)
...
+			return GetCellBounds (row, cell, includeMargin);
+		}
+
+		public Rectangle GetRowBounds (TreePosition pos, bool includeMargin)
+		{
+			var it = tsource.GetItem (pos);
+			if (it == null)
+				return Rectangle.Zero;
+			var row = Tree.RowForItem (it);
```
slluis: The result of Tree.RowForItem has to be casted to int, since it returns an nint on XamMac.

sevoku: Ok, and what about other RowForItem calls ([Line 148](https://github.com/mono/xwt/pull/433/files#diff-3e118f34da2c8971c2c95f0591537432R148) or [Line 154](https://github.com/mono/xwt/pull/433/files#diff-3e118f34da2c8971c2c95f0591537432R154))? Shouldn't it always be casted then?

Xwt.XamMac/Xwt.Mac/TreeViewBackend.cs
sevoku: Ok, I have Xamarin.Mac now and everything compiles.

sevoku:
@slluis: I've checked the bounds on Mac one more time and was able to reproduce the problem: the bounds are correct (simply dump them to the console). The problem seems to be in MacEngine.RenderWidget when the rendered widget is focused. While the bounds remain the same, the widget coordinates inside the rendered image change. To reproduce this simply focus some other widget (e.g. list with the samples on the left) and move the mouse over the sample list again. The screenshot of the row should be correct now. So this is some other issue, which needs some investigation, but has nothing to do with the row bounds implementation.

PR #435, opened 2015-01-07
[Wpf] Fix CanvasCellView Events and Redraw

sevoku:
This PR adds event handling to CellViewBackend and the according implementation for CanvasCellView (other cell view types still need to be implemented). Fixes additionally CanvasCellViewBackend.QueueDraw().

CustomCell Example (ListView1.cs) works now with Wpf :grin:


sevoku:
See #531 for a better approach.

PR #439, opened 2015-01-10
Finding TreeNavigators by value in TreeStore.

residuum:
This traverses the TreeStore to find the TreeNavigator of the first item that contains a specified value in the data field.

One use case: Getting the TreePosition of the navigator and then select the corresponding row.


residuum:
There may be more than one item in the TreeStore that contains that value. Therefore, the method now returns a `IEnumerable<TreeNavigator>` and was renamed accordingly.

residuum:
Is there a reason, why this has not been merged yet?

Therzok:
This looks good to me.

PR #445, opened 2015-01-16
[Mac] fix button sizing and colors

sevoku:
*Update: fixes button colors*

**Button sizing:**
The default NSBezelStyle.Rounded style (which is the default for text buttons on Mac)  does not support cusom heights.
When a custom size or image is used, the style should be set to NSBezelStyle.RegularSquare (the default for image or custom buttons on Mac) for the sizing to work as expected.

### DIFF
--- a/Xwt.Mac/Xwt.Mac/ButtonBackend.cs
+++ b/Xwt.Mac/Xwt.Mac/ButtonBackend.cs
@@ -36,6 +36,9 @@ namespace Xwt.Mac
 {
 	public class ButtonBackend: ViewBackend<NSButton,IButtonEventSink>, IButtonBackend
 	{
+		ButtonType currentType;
+		ButtonStyle currentStyle;
+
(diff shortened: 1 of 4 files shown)


sevoku:
Some screenshots:
**Before:**
xwt mac button sizing issue

**After:**
xwt mac button sizing fixed

sevoku:
**Button Colors:**
When setting BackgroundColor on an NSButton, only the background layer is colored and not the button (Bezel) itself. The button has always the default system colors (blue if rounded, grey if squared). To fix this, the view color is left untouched and the button is rendered to an NSImage and a [color matrix filter](https://developer.apple.com/library/mac/documentation/GraphicsImaging/Reference/CoreImageFilterReference/index.html#//apple_ref/doc/filter/ci/CIColorMatrix) is applied.

**Before:**
xwt mac button colors
**After:**
xwt mac button colors fixed

PR #449, opened 2015-01-16
Consistent overloads for TreeNavigator.SetValues().

residuum:
The parameter `int row` is not used in the overloads, and only the overloads for 2 and 3 values use it.

In order not to break existing code, the original signatures are still there, but call the new methods and are marked with `ObsoleteAttribute`.


slluis:
Let's just remove the row argument. It is ok to break the api, we are not yet frozen.

residuum:
Ping.

sevoku:
Thanks!

PR #450, opened 2015-01-17
Font optimization

sevoku:
While working on the FontSelector widget I've found some issues with the current font implementation. 

1. missing weight values (a84e9ff): according to [Pango API](https://developer.gnome.org/pango/stable/pango-Fonts.html#PangoWeight) there are some additional predefined weight values and they have backend specific (Wpf/Mac) pendants. So I have added them and optimized the backend mappings for better fitting.
2. little Mac issue (016c67a): the trait mask has to be updated, when changing the font weight (NSFontTraitMask.Bold, NSFontTraitMask.Unbold)
3. A new function to retrieve available font family variants (c3000f6): since the Font object allows to apply any weight, style and stretch combination (even invalid), I was missing a function to get really available/valid combinations (font/system specific, of course). ```GetAvailableFontFaces``` returns a collection of face specific names and their Font representations.

### DIFF
--- a/Xwt.Gtk/Xwt.GtkBackend/FontBackendHandler.cs
+++ b/Xwt.Gtk/Xwt.GtkBackend/FontBackendHandler.cs
@@ -24,12 +24,10 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-using System;
 using Xwt.Backends;
 using Pango;
 using Xwt.Drawing;
 using System.Globalization;
-using System.Reflection;
 using System.Collections.Generic;
 using System.Linq;
 
@@ -37,6 +35,20 @@ namespace Xwt.GtkBackend
 {
 	public class GtkFontBackendHandler: FontBackendHandler
 	{
+		static Pango.Context systemContext;
+		static Dictionary<string, Dictionary<string, Pango.FontFace>> familyFontFaces;
+
+		static GtkFontBackendHandler ()
+		{
+			systemContext = Gdk.PangoHelper.ContextGet ();
+			familyFontFaces = new Dictionary<string, Dictionary<string, Pango.FontFace>> ();
+			foreach (var family in systemContext.Families) {
+				familyFontFaces [family.Name] = new Dictionary<string, Pango.FontFace> ();
+				foreach (var face in family.Faces)
+					familyFontFaces [family.Name] [face.FaceName] = face;
+			}
+		}
+
 		public override object GetSystemDefaultFont ()
 		{
 			var la = new Gtk.Label ("");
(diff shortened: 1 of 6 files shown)


Xwt.Gtk/Xwt.GtkBackend/FontBackendHandler.cs
```diff
@@ -45,7 +57,15 @@ public override object GetSystemDefaultFont ()
...
 		public override IEnumerable<string> GetInstalledFonts ()
 		{
-			return Gdk.PangoHelper.ContextGet ().FontMap.Families.Select (f => f.Name);
+			return systemContext.FontMap.Families.Select (f => f.Name);
+		}
+
+		public override IEnumerable<KeyValuePair<string, object>> GetAvailableFamilyFaces (string family)
+		{
+			if (familyFontFaces.ContainsKey (family))
+				foreach (var fd in familyFontFaces [family].Values)
```
slluis: Use TryGetValue here to avoid unnecessary dictionary access

Xwt.Gtk/Xwt.GtkBackend/FontBackendHandler.cs
```diff
@@ -45,7 +57,15 @@ public override object GetSystemDefaultFont ()
...
+					yield return new KeyValuePair<string, object>(fd.FaceName, fd.Describe ());
```
slluis: why not cache the whole KeyValuePair in familyFontFaces?

sevoku:
Ok, I've optimized the font caching in the Gtk backend. Do we need some caching in the other backends, too?

slluis:
Hmm, thinking a bit more about this, I'm wondering what's the memory impact of having all those font descriptions loaded in memory. If it is noticeable, maybe font face and font family information should be loaded and cached on demand.

sevoku:
Well, on my Ubuntu there are something about 160 fonts installed. So there are something about 40-50 base fonts. This should be more or less the normal situation.  But on some systems used by designers this could be much more, since they often use many different custom fonts. So caching on demand could increase the performance and memory usage a little bit.

And the current solution has a little problem with newly installed fonts. They will appear only after an application restart. So the caching needs some more code.

sevoku:
How about half caching (only cached PangoDescription) like this:
```c#
public override IEnumerable<KeyValuePair<string, object>> GetAvailableFamilyFaces (string family)
{
	Dictionary<string, object> fontFaces;

	FontFamily pangoFamily = systemContext.Families.FirstOrDefault (f => f.Name == family);
	if (pangoFamily == null)
		yield break;

	if (!familyFontFaces.TryGetValue (family, out fontFaces))
		familyFontFaces[family] = fontFaces = new Dictionary<string, object> ();

	foreach (var ff in pangoFamily.Faces) {
		object fd;
		if (!fontFaces.TryGetValue (ff.FaceName, out fd))
			fontFaces [ff.FaceName] = fd = ff.Describe ();
		yield return new KeyValuePair<string, object>(ff.FaceName, fd);
	}
	yield break;
}
```
without the initialization in the static constructor then.

sevoku:
I have removed the caching completely. It doesn't really make sense to cache only the font variants but not all fonts (and only for Gtk). Perhaps we should think about some generic font caching for all backends (maybe interesting for the discussion in #461, too).

slluis:
It looks good to me now.

sevoku:
Should we merge it now, or better later for 0.3 (because it changes the API slightly: new weight values and different weight backend mapping)?

slluis:
I think it is ok to commit it now.

PR #455, opened 2015-01-18
[Wpf] Fixing unit tests crash + TreeView.HiddenTree test

DavidKarlas:
What this code does... It forces loading of vitalized TreeViewItem so SelectItem can be performed... Before exception was throw because treeViewItem was null and unit tests running crashed...

### DIFF
--- a/Xwt.WPF/Xwt.WPFBackend/ExTreeView.cs
+++ b/Xwt.WPF/Xwt.WPFBackend/ExTreeView.cs
@@ -37,6 +37,7 @@ using System.Collections.Generic;
 using SWC = System.Windows.Controls;
 using WKey = System.Windows.Input.Key;
 using System.Windows.Input;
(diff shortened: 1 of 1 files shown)


slluis:
I don't know enough of WPF to validate this.

DavidKarlas:
It's not well documented either. Since IItemContainerGenerator(notice double I in front) is mostly used by WPF itself internally. This interface is used to request generation of TreeViewItem(UI representation of data entry). What this code does... If g.ContainerFromItem fails to get value it requests generation of first ungenerated treeViewItem and then tries again to get that item. I can't think of any regression created by this. I just wanted to fix WPF unit tests crash.

sevoku:
I think we can merge this. I'm not really good in WPF, too. But as long the tests pass and there are no obvious regressions, we should take it. Otherwise someone can create a new issue.

PR #461, opened 2015-01-19
Allow TextLayout and Font to be re-initialized with a different toolkit

garuma:
Also change Mac's ContextBackendHandler to request a safe version of TextLayout to cope with layout created on a different toolkit.

This tries to mimic the way Image on foreign-toolkits are handled.


sevoku:
Is this required on Gtk/Wpf, too? Both use GetBackend and not GetSafeBackend.

garuma:
I think the classic case (at least the one we have) is to go from object originating from Gtk+ to a Mac context. Could be added to the other backend for the sake of completeness though.

garuma:
I added Font transformation through the same mechanism. Also added call to GetSafeBackend to Wpf backend too.

Xwt.Mac/Xwt.Mac/ContextBackendHandler.cs
```diff
@@ -276,7 +276,8 @@ public override void DrawTextLayout (object backend, TextLayout layout, double x
 		{
 			CGContext ctx = ((CGContextBackend)backend).Context;
 			SetupContextForDrawing (ctx);
-			MacTextLayoutBackendHandler.Draw (ctx, Toolkit.GetBackend (layout), x, y);
+			var li = ApplicationContext.Toolkit.GetSafeBackend (layout);
```
slluis: This should be handled by the frontend. DrawTextLayout should take a TextLayout backend (object) instead of directly a TextLayout, so the caller will do the GetSafeBackend() call and there will be no need for platform-specific checks.

Xwt/Xwt/Toolkit.cs
```diff
@@ -375,6 +375,10 @@ public object ValidateObject (object obj)
 		{
 			if (obj is Image)
 				((Image)obj).InitForToolkit (this);
+			else if (obj is TextLayout)
+				((TextLayout)obj).InitForToolkit (this);
+			else if (obj is Font)
+				return ((IFrontend)obj).ToolkitEngine != this ? Font.FromName (obj.ToString (), this) : obj;
```
slluis: This is not very efficient. If a Font is created in for the GTK toolkit and then used intensively in Mac it will result in fonts being constantly recreated.

Xwt.Mac/Xwt.Mac/ContextBackendHandler.cs
garuma: Sure although that's a more thorough change through the codebase so how do you want to proceed with this?

Xwt/Xwt/Toolkit.cs
garuma: Only place that calls ValidateObject for a font right now is `TextLayout` because I'm assuming that's the only place that will in my situation (foreign toolkit drawing).  Because of that, that call would only be used once via the `InitForToolkit` check of `TextLayout`.

Originally I had that code directly in the `TextLayout` class, I can move it back there if you want. Otherwise should we go for the same `InitForToolkit` style of things?

slluis: Yes, we should go with a InitForToolkit method.

Xwt.Mac/Xwt.Mac/ContextBackendHandler.cs
slluis: Let's keep it as is for now.

garuma:
I think we are good on this, permission to merge?

sevoku:
How about Gtk and GetSafeBackend?

garuma:
There you go

Issue #53, opened 2012-03-15
Toolbar implementation missing


zaubererty:
There is currently  no front and back end implementation for a toolbar I'm pleased to help out with some guidance.

sevoku:
Any progress here?

dlech:
I never got very far with this. The relevant branches for the commands I mentioned are <https://github.com/dlech/xwt/tree/commands_v1> and <https://github.com/dlech/xwt/tree/commands_v2> if any of that is useful.

cra0zy:
I want to try and setup this. Few questions about toolbars:

1) Do Gtk HeaderBar:
![head](https://cloud.githubusercontent.com/assets/6773302/8462694/d042ce8c-2035-11e5-8dc2-01ed56de9263.png)
and Mac's Window toolbar:
![head](http://i.stack.imgur.com/bZCVd.png)
count as toolbars?

2) Should toolbar act as MainMenu? (ie. just one toolbar per window)

3) Should there be maybe 2 different toolbar implementation, 1 for headerbar / mac integrated toolbar that works like MainMenu, and 1 for normal toolbars?

4) Can you put whatever widget you want in Mac toolbar(I already know that Gtk and Wpf can put whatever widget you want inside their toolbars)?

Therzok:
1. Imo, yes.
4. Yes. But you have to manually manage the NSToolbarItem sizes.

sevoku:
1. I don't think so. "Classic" Toolbars are much more flexible. The header bars should provide only some main functions, maybe context based additionally. Where Toolbars are more customizable and an application can have more then one (e.g. office applications).

2. A MainMenu is something different. I think a HeaderBar is can exist additionally to menus or toolbars. But one can be part of an other, I think. So it should be possible to put a main menu inside a header bar, and so on.

3. In my opinion yes. There should be a Toolbar, a MainMenu and a HeaderBar. Its up to the context and the app design to implement all of them or only what is required.

4. It would be really cool, if it were possible to implement header bars on all platforms! It should be a simple HBox container, which can hold any type of widgets, or menus.

PR #463, opened 2015-01-19
Commands with writable properties

sevoku:
Even if implemented (e.g. IAlertDialogBackend) the command icons are never assigned, since private. This PR lets the user set custom command icons and labels.

This is connected to #53, where a new Command concept is proposed. Until we have some more sophisticated solution, this is an easy way to allow icons on dialog buttons, or create menu items using commands.

### DIFF
--- a/Xwt/Xwt/Command.cs
+++ b/Xwt/Xwt/Command.cs
@@ -30,10 +30,23 @@ namespace Xwt
 {
 	public class Command
 	{
-		public Command (string id)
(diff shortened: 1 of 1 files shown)


slluis:
The constructors look good, but Command should be immutable, so it should not have writable properties.

sevoku:
Ok! I wanted to make it possible to change the default commands, but this can be solved with the new command concept.

PR #471, opened 2015-01-21
[GTK] fix resize artifacts, reuse popover window

sevoku:
The Gtk PopoverBackend can reuse the popover window and destroy it on Dispose, instead of creating a new window each time the Popover is shown.


sevoku:
The last commit (f24c449) fixes the visual corruption, when the popover size changes:
gtktest_039  gtktest_040

PR #472, opened 2015-01-21
[Gtk] paned events implementation

sevoku:
I have cherry-picked relevant parts from #186 (f85e2ad601bfafe7da3c55be6b419ff31b1f4ad8 and 357714ef4b757b358daaa1e34f89b0562c8f92e8) and replaced the WidgetEvent subscription with a GLib notification to reduce event overhead.

### DIFF
--- a/Xwt.Gtk/Xwt.GtkBackend/PanedBackend.cs
+++ b/Xwt.Gtk/Xwt.GtkBackend/PanedBackend.cs
@@ -106,11 +106,34 @@ namespace Xwt.GtkBackend
 			throw new NotSupportedException ();
 		}
 		
+		public override void EnableEvent (object eventId)
+		{
+			base.EnableEvent (eventId);
+			if (eventId is PanedEvent) {
+				switch ((PanedEvent)eventId) {
+				case PanedEvent.PositionChanged: Widget.WidgetEvent += HandleWidgetEvent; break;
+				}
+			}
+		}
+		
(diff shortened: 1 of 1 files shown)


sevoku:
@slluis: is a cherry-pick ok? The commit author is left untouched. I only added more descriptive messages.

Xwt.Gtk/Xwt.GtkBackend/PanedBackend.cs
```diff
@@ -92,15 +88,34 @@ public void UpdatePanel (int panel, bool resize, bool shrink)
...
+						Widget.RemoveNotification ("position", HandleMove);break;
+				}
+			}
 		}
 
-		public void GetPanelSizes (double totalSize, out double panel1Size, out double panel2Size)
+		void HandleMove (object o, GLib.NotifyArgs args)
 		{
-			throw new NotSupportedException ();
+			ApplicationContext.InvokeUserCode (delegate {
```
slluis: Shouldn't the property name be checked here?

sevoku: The handler is connected to the "position" property change event only (notification subscription in [Line 98](https://github.com/mono/xwt/pull/472/files#diff-58e52c205c8806a597fecc8b172941e3R98)). So the check would be redundant. A property name check would make sense, if the handler would handle notifications for more then one property.

PR #473, opened 2015-01-22
[Gtk] fix SetFocus

sevoku:
This is based on #310. As discussed ```Widget.IsFocus = true``` is not always enough. Some Gtk widgets (ListView, maybe others) need sometimes a GrabFocus() to receive keyboard focus and those which don't need it (should) behave as expected/before. Additionally an example is included.

Note: I've removed ```void SetFocus (Gtk.Widget w)``` which is never used.


slluis:
We really need unit tests for all of this.

sevoku:
There are Focus tests and they pass with this. Or do you mean something special?

sevoku:
Oh, well, we need some deeper Focus tests. For Lists we should test if a row is focused, and so on. But this is much work...

PR #475, opened 2015-01-27
[Gtk] Optimize SearchTextEntry (and TextEntry.PlaceholderText rendering)

sevoku:
This PR replaces the complex SearchEntry implementation (seems to have been copied from MonoDevelop) with a native Gtk.Entry implementation using the native Icon feature of Gtk 2.16+.

Benefits:
* No complex code using EventBoxes, custom image drawers, etc.
* same backend for Gtk2 and Gtk3
* uses theme icons (search, clear) with custom fallback icons

Regressions:
* does not work as expected with Gtk 2.19 or lower (Gtk produces warnings, icons not visible, but no exceptions)

Thats how it looks like (screenshots from MonoDevelop NuGet Packages Dialog on Ubuntu):
Before:
add packages_044
After:
add packages_042


Xwt.Gtk/Xwt.GtkBackend/Gtk2IconEntry.cs
```diff
@@ -0,0 +1,264 @@
...
+}
+
+namespace Gtk
+{
+	public delegate void IconPressHandler(object o, IconPressArgs args);
+
+	public delegate void IconReleaseHandler (object o, IconReleaseArgs args);
+
+	public class IconPressArgs : GLib.SignalArgs {
+		public Gtk.EntryIconPosition P0{
```
Therzok: Some styling issues for these properties and the ones below?

sevoku: It's from the Gtk generator, to make Gtk2 and 3 work the same way. The long mission is to add customizable icons for TextEntry.

sevoku: An other way could be to add these features to upstream gtk-sharp, but Xwt would be only compatible with the most current gtk-sharp version then. So not really a good solution. Sure this implementation won't win a code style competition, but its still better then the really large MD variant, where many features are not even used within Xwt.
Even more, the MD search entry looks alien on all platforms (I mean the one in the main ui, which the original Xwt implementation uses atm) to me. The entry has a different geometry then other entries and the icons have different alignments on different platforms, so it doesn't really hit the aim of Xwt to make all widgets native looking.

sevoku:
I've removed the ugly generator code, so no more Gtk namespace inside Xwt. GLib requires enums to be registered, so I register a private enum and use only the Args array inside the handler by casting to int. I think this is a cleaner solution and should not interfere with the gtk-sharp API, if the icons will be added to v2 some day.

@Therzok: what do you think? This way (https://github.com/sevoku/xwt/blob/a8127b1739024677362d5868d2abda7df1f3acf6/Xwt.Gtk/Xwt.GtkBackend/Gtk2IconEntry.cs) it could be mergable?

Therzok:
This looks good to me.

sevoku:
Thanks for the review! I'll test it with MD master tomorrow and merge if everything works correctly.

sevoku:
I have a little update on this: it works, but not very good with MD/XS :(

The main problem is the lack of transparency in Gtk2 and has nothing to do with this special case. Gtk.Entry has its own Gdk.Window which is used for drawing. Since Gtk2 has no transparency support this Gdk.Window is always white or filled by Gtk with the default bg color. So if an entry is placed inside some custom widget/canvas, it has always a little grey border (outside the real themed border, in most cases with rounded corners). This can be reproduced in the canvas sample.

XS/MD uses this SearchEntry inside the header bar of the NuGet package dialog, which IS a canvas.
The original implementation is a special workaround and uses an EventBox with an aligned frameless entry inside. The border is drawn manually outside the entry on the eventbox, which is transparent.

So replacing the original implementation with the native one brings the thin grey border to the NuGet dialog used in XS/MD.

original MD on Mac: bildschirmfoto vom 2015-07-26 18-58-08
native (this PR) on Mac: bildschirmfoto vom 2015-07-26 18-57-10

I think a more consistent solution is needed: 
1. Merge this PR for a native impl. in Xwt
2. Add a workaround to MD (e.g. create a custom canvas widget replacing the Gtk impl)
3. Find a working fix for all Gtk.Entry based Xwt widgets (TextEntry, SearchTextEntry, ComboBox, ComboBoxEntry)

Issue #470, opened 2015-01-21
Unified StockIcons solution


sevoku:
Since we are still missing many stock icons and the declared ones are not available on all backends we should think about a unified solution for this.

Current state: Wpf and Mac have only the small system sets. Gtk has much more, but the stock-item things have been deprecated (and will be removed some day).

One solution could be to remove stock icons completely, since we'll never find more matching system icons on Wpf/Mac (or they change their design concept).

An other way could be to add backend independent icons, what I personally would prefer. We could include the full [Adwaita](https://github.com/GNOME/adwaita-icon-theme) icon set (the icons already included, like list-add.png in Wpf, seem to come from Adwaita). And Adwaita has a compatible [CCBYSA3](http://creativecommons.org/licenses/by-sa/3.0/) license (a reference should be added anyway, if we keep the current state!)

Additionally StockIcon override methods could be implemented. Something like ```Application.SetStockIcon(string id, Image icon)``` and ```Application.SetStockIconFallback(string id, Image icon)``` which can be used to set a fallback icon (used in case the backend has no matching icon) or to completely override the backend.

slluis:
I think we should remove the stock icons, or reduce them to a very limited set. The original idea was to provide icons for common commands and for message dialogs, but I don't think that's very useful anymore. I also don't think it is a good idea to use a platform-independent icon set, because it will look weird in all platforms.

sevoku:
I agree, one default set for all, would look strange. But there are some common icons that are very often used and its more simple to load a stock icon (current API) then to load one from a resource (unique enum id vs. string). And I wouldn't drop the original idea, since it will be much more useful for a Toolbar backend (some day).
The current set is already (or still?) very limited. And I think we can't simply remove StockIcons, since at least the MessageDialogs would not work without them (we could make them internal, though). So why not extend it with more common icons (like file, folder, etc) and add the proposed functions (an ```Application.SetStockIcons(IEnumerable<StockIconId id, Image icon> inconSet)``` would be nice too, to override many icons at once, and we could add a function for testing if a toolkit provides a special stock icon)? To avoid strange looking apps, we could really remove wrong icons (wpf has some) and throw the NotImplemented exception (like before, maybe with a better description) without adding one default set.

It could be a round thing, if it we add good documentation about missing icons and how to add them. And the existing ones make an app really more native looking.

To summerize the idea:
* Add more useful stock ids
* remove all custom icons from code (Wpf/Mac) and leave those loaded from the system
[...]

PR #477, opened 2015-01-28
Iconeria: resolution independent icons 

lytico:
including  http://fortawesome.github.io/Font-Awesome/
(Version 4.04)
usage:
var icons = new AwesomeIconeria {Fill = true,FillColor = Colors.White,
							Stroke = true,StrokeColor = Colors.Black};
// make an icon with 36 pixels:
var img = Iconeria.AsImage (icons.FaThumbsUp, 36);

Background:
AwesomeIconeria (see file: AwesomeIconeria-4.04.Generated.cs) has a method for each icon. This methods have been gererated with a tool (https://github.com/lytico/svgpath2code/tree/MonoXwtVersion) out of the font-definitions of Font Awesome.


sevoku:
I don't really understand how this works. We have to add a drawing method for every stock icon (marked with the IconAttribute)? Or is it a way for the user to implement own icons and why should this be part of Xwt then?

lytico:
Yes, there IS a drawing method for each font awesome icon, see file:
AwesomeIconeria-4.04.Generated.cs
You have do click SHOW, as it is not expanded in the standard view of github, cause its too big

sevoku:
Aaah! Sorry, I didn't see that! Nice!
But I'm still not sure about adding this to Xwt itself. If there would be a project for different platforms, then this could be added there, like I did it for [Oxyplot](http://oxyplot.org/announcements/). But it looks like there are tons of different specialized projects (Wpf, Xamarin,...) on [NuGet](http://www.nuget.org/packages?q=FontAwesome). We could create a new separate Project (something like FontAwesome.Xwt) and bind it to an Xwt NuGet package (when it gets released).

lytico:
It could be with more icon sets, they only need to have svgpath definition. eg. https://github.com/GNOME/adwaita-icon-theme (as you suggested in https://github.com/mono/xwt/issues/470) would be possible too, it needs only adoption of the svg to xwt-code converter.
With project you mean github project, not c#-project, I suppose?

sevoku:
Concerning #470: thats the point. The idea behind StockIcons is to use native icons (which come from the platform itself and can change depending on platform release (Gnome 2 or 3, Windows 7/Windows 8,...) and user settings (e.g. Gtk Themes). My idea is to make them replaceable from the user side.

Your proposal is a really nice way for the user to generate custom static icons, which is independent of #470 and the idea behind Xwt or StockIcons. So yes I really mean a separate development project (on GitHub or wherever), which could include a generator for custom icon sets and more. @slluis: what do you think?

Generator: how does it work? Do you have some code? Btw: I'd like to implement Svg support for Xwt (load images from svg using a special backends like rsvg with Gtk) but this is an other story...

lytico:
Generator: https://github.com/lytico/svgpath2code/tree/MonoXwtVersion
its a modified version of https://github.com/spouliot/svgpath2code . The problem is the license, its LGPL2, so it can't included in core xwt, otherwise it would be a good base for svg support

sevoku:
Hhm, without a generator, it doesn't really make sense to add only this special icon set to Xwt. An own project should be the way to go, which should be LGPL licensed then. We could add a "Projects using Xwt" section to the Xwt readme.md with some links to projects like this one :+1:

slluis:
I agree with Vsevolod, this doesn't really belong to XWT, it should be in its own repo.

slluis:
However, let me add that this is very cool :)

sevoku:
For sure! Really cool!

sevoku:
I think we can close this. Let us know if you need some help with the project!

Issue #257, opened 2013-11-14
Window position cannot be properly set


ksigne:
This code
<code>
Xwt.Window W = new Xwt.Window();
W.X = W.X;
W.Y = W.Y;
</code>
Makes window move to a random direction. It should be caused by inproper coordinate transformation.

sevoku:
Any progress on this?

sevoku:
I've added a PR that hopefully fixes this for Gtk: #479
But i think we have an other issue with Window Bounds and Positions in: #480

PR #479, opened 2015-01-29
[Gtk] fix window position

sevoku:
This is a fix for #257 including tests. The problem was, that the Gtk.Window has been used for setting the location/size and its Gdk.Window to retrieve the Bounds. Since Gtk.Window always includes the Decorations and its Gdk.Window (used for drawing) includes only the client side (without decorations), the decoration size was always added. 

I've included tests for location changes and preserving location (as described in #257). 
* WPF: only does not pass LocationChangeWithoutContent
* MAC: does not pass at all.
* GTK: passes all tests

### DIFF
--- a/Testing/Tests/WindowTests.cs
+++ b/Testing/Tests/WindowTests.cs
@@ -256,6 +256,77 @@ namespace Xwt
 				Assert.AreEqual (230, win.Y);
 			}
 		}
+
+		[Test]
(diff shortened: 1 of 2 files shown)


sevoku:
Sorry, the bounds should **include** the decorations. But with this change Gtk does not pass the Coordinates tests (like no other backend, see #480).

PR #481, opened 2015-01-30
[WPF] Calling TreeStore.Clear() raises exception in ExTreeViewItem

residuum:
When one or more top nodes in TreeView are expanded, calling `Clear()` on the ITreeDataSource raises an exception.

This is due to `OnCollapsed (RoutedEventArgs e)` being called in ExTreeViewItem, and this cast fails:

    var node = (TreeStoreNode)DataContext;

Specifically: `DataContext` is `DisconnectedItem` of type `MS.Internal.NamedObject`

### DIFF
--- a/Xwt.WPF/Xwt.WPFBackend/ExTreeViewItem.cs
+++ b/Xwt.WPF/Xwt.WPFBackend/ExTreeViewItem.cs
@@ -65,7 +65,10 @@ namespace Xwt.WPFBackend
 
 		protected override void OnCollapsed(RoutedEventArgs e)
 		{
-			var node = (TreeStoreNode)DataContext;
+			var node = DataContext as TreeStoreNode;
+			if (node == null) {
+				return;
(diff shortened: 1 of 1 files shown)


Xwt.WPF/Xwt.WPFBackend/ExTreeViewItem.cs
```diff
@@ -65,6 +65,9 @@ protected override void OnExpanded (RoutedEventArgs e)
 
 		protected override void OnCollapsed(RoutedEventArgs e)
 		{
+			if (!(DataContext is TreeStoreNode)) {
```
sevoku: Looks right, but you should use `as` and check against null.
```c#
var node = DataContext as TreeStoreNode;
if (node == null) return;
```

residuum:
done.

sevoku:
Thanks!

PR #483, opened 2015-02-02
Handle an invisible TextCellView in WPF

steffenWi:
Since there is a 'Visible' Property in TextCellView (and I'm using it so that I can place a hidden GUID on each row that helps me associate a row with an object) the property should be respected.

I'm not sure if this is possible/available in GTK or Cococa, but it does work in WPF.

I'm currently using it by creating a ListViewColumn with a 'normal', visible TextCellView inside. Later I add the invisible TextCellView to the Views collection of the ListViewColumn.

MaxWidth has to be 0 because the column resizes so that both (all?) TextCellViews are accounted for.

### DIFF
--- a/Xwt.WPF/Xwt.WPFBackend.Utilities/CellUtil.cs
+++ b/Xwt.WPF/Xwt.WPFBackend.Utilities/CellUtil.cs
@@ -76,6 +76,12 @@ namespace Xwt.WPFBackend.Utilities
 					{
 						factory = new FrameworkElementFactory(typeof(SWC.TextBlock));
 						factory.SetValue(FrameworkElement.MarginProperty, CellMargins);
+						if (!view.Visible)
+						{
+							factory.SetValue(FrameworkElement.VisibilityProperty, Visibility.Hidden);
+							factory.SetValue(FrameworkElement.MaxWidthProperty, 0.0);
+						}
+
 						if (textView.TextField != null)
 						{
 							factory.SetBinding(SWC.TextBlock.TextProperty, new Binding(dataPath + "[" + textView.TextField.Index + "]"));
--- a/Xwt.WPF/Xwt.WPFBackend.Utilities/CellUtil.cs
+++ b/Xwt.WPF/Xwt.WPFBackend.Utilities/CellUtil.cs
@@ -76,11 +76,14 @@ namespace Xwt.WPFBackend.Utilities
 					{
 						factory = new FrameworkElementFactory(typeof(SWC.TextBlock));
 						factory.SetValue(FrameworkElement.MarginProperty, CellMargins);
-						if (!view.Visible)
+						if (textView.VisibleField != null)
 						{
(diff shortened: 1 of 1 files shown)


sevoku:
What happens, if the property is changed, or if the user sets VisibleField? Can't a Binding be used, like for the Text property?

Btw: if you only need a field to store a GUID which is always invisible, you do not need to add a view for that field.

steffenWi:
Maybe. I oriented myself on how the MarginProperty is set.

On the second part of your comment: I'm not sure about that. A ListViewColumn needs some kind of CellView and the 'default' CellView is not supported by the XWT WPF implementation (and I'm not sure how to implement it).

sevoku:
You don't even need to add a coulumn. Not all fields in a store must have a matching column. So you can leave both out (column and view).

sevoku:
Ok, I had some minutes to take a deeper look on it :)

1. As I have written, if you need a "row id", simply add a DataField<Guid> to your store. You don't need to add any column or view to your ListView/TreeView for that field, if you don't want the field to be ever visible. It makes only sense if you want to show and hide a CellView dynamically (using the Visible property for all rows, or using a separate DataField<bool> assigned to VisibleField to show/hide row wise).

2. I don't really understand why you use Visibility.Hidden and limit the width. Why don't use Visibility.Collapsed (which does not allocate any space for the view according to https://msdn.microsoft.com/en-us/library/system.windows.visibility)?

The right way to implement visibilitiy should be:

```c#
if (textView.VisibleField != null) {
	var binding = new Binding (dataPath + "[" + textView.VisibleField.Index + "]");
	binding.Converter = new BooleanToVisibilityConverter ();
	factory.SetBinding (SWC.TextBlock.VisibilityProperty, binding);
} else if (!textView.Visible)
	factory.SetValue (SWC.TextBlock.VisibilityProperty, Visibility.Collapsed);
```
This takes CellView.Visible and CellView.VisibleField into account. If setting MaxWidth=0 is really required, the a ValueConverter should be implemented to convert the width.

steffenWi:
As to 1st: I still don't understand how I'd add a DataField to the ListView without creating a column?! ListView.Columns.Add only takes a ListViewColumn, which in turn needs a CellView of some kind which takes the DataField. Or do you mean that I should just add stuff to the ListView's DataSource? I'm pretty sure I tried that and got an IndexOutOfRange exception.

On the second part: That should work (and makes more sense).

sevoku:
Yes, you should be able to add it to the DataSource only. If you get an exception while accessing the (unbound) DataField, then there must an other bug somewhere.

steffenWi:
okay, on the topic of binding: Your changes do work. How do I go from here? Should I close this PR and open a new one after doing a reset HEAD^ on my branch?

steffenWi:
and you were right again: Just adding the DataField to the ListStore works. Nothing else needed to be changed.

steffenWi:
and done :)

PR #486, opened 2015-02-06
Make stock icons settable.

residuum:
Inspired by pull request #477, I have started a simple solution to create icon sets.

This is just adding an interface to implement, and a way to set that class to the StockIcons class. If an icon set returns `null` for an StockImageId, then the stock icon of the toolkit is used.

This will make it simpler to use your own icons in MessageDialog.ShowMessage() et al.

### DIFF
--- a/Xwt.WPF/Xwt.WPFBackend/AlertDialogBackend.cs
+++ b/Xwt.WPF/Xwt.WPFBackend/AlertDialogBackend.cs
@@ -54,8 +54,9 @@ namespace Xwt.WPFBackend
 
 		public Command Run (WindowFrame transientFor, MessageDescription message)
 		{
-			this.icon = GetIcon (message.Icon);
-			if (ConvertButtons (message.Buttons, out buttons)) {
+			this.icon = GetIcon(message.Icon);
+			bool isCustomIcon = message.Icon != null && this.icon == MessageBoxImage.None;
+			if (ConvertButtons (message.Buttons, out buttons) && !isCustomIcon) {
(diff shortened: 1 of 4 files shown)


residuum:
The name `StockIcons` for the static class is not correct any more after the merge, though.

sevoku:
This is a nice idea, but the changes look to specific to me (specific support in dialogs). May I point you to #470? I think there must be a way to replace icons on the deeper backend level. I've proposed something like `SetStockIcons(IEnumerable<StockIconId id, Image icon> inconSet, bool overrideToolkit = false)` which would override an icon before it is loaded by the backend, so no change to e.g. the dielog backend implementation would be needed.

residuum:
That is OK with me, if there is a different solution for overriding the stock icons.

I was just using `MessageDialog.ShowMessage ()` and have been wondering, how to set my own icons instead of having to use the default icons without the need to use my own windows.

Everywhere else, overriding the stock icons is easy, just replace every usage of `StockIcons.Add` with `MyIconClass.Add` and you are done.

sevoku:
We have introduced a new `Toolkit.Defaults` property in https://github.com/mono/xwt/commit/5b9f4545ebb3a3faf771b34fb4ef2187645001b8. For now there is only a `MessageDialog` member which allows you to set custom dialog icons, this may be extended in the future.

@slluis without the WPF part, I like the idea of the StockIcons.IconSet, what do you think?

residuum:
Thanks, that does it for me.

PR #488, opened 2015-02-11
Use Dispatcher to Dispose of a Window

steffenWi:
... if the Dispose was called from a seperate thread than the window belongs to.

### DIFF
--- a/Xwt.WPF/Xwt.WPFBackend/WindowFrameBackend.cs
+++ b/Xwt.WPF/Xwt.WPFBackend/WindowFrameBackend.cs
@@ -28,8 +28,9 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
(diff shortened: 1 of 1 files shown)


steffenWi:
Hi,

why is this not merged? Are there any problems with my code?

sevoku:
The code is ok! Works for me. But the commit history is confusing. Can you rebase so it is one single commit containing the relevant change only, please?

steffenWi:
history fixed and formatting in extra commit

steffenWi:
@sevoku hey, does this look ok to you?

steffenWi:
@sevoku you there?

Therzok:
The problem is that we do not use VS style as a coding style in Xwt. I opened https://github.com/mono/xwt/pull/558/files with just the functional change.

PR #489, opened 2015-02-13
Test fixes

sevoku:
I'd like to collect some test fixes in the [test-fixes](https://github.com/mono/xwt/tree/test-fixes) branch for the upcoming CI integration and NuGet builds.

To propose more fixes, please create a PR against the [test-fixes](https://github.com/mono/xwt/tree/test-fixes) branch.


sevoku:
We provide NUnit binaries for the tests, shouldn't we remove them and add a git submodule for NUnit?

slluis:
We should use the NUnit NuGet

Issue #286, opened 2014-02-02
RowActivated is not tracked in WPF


ksigne:
ListView.RowActivated has no implementation in WPF backend.

PR #490, opened 2015-02-13
[WPF] ListBox and ListView: add support to RowActivated event

henriquemotaesteves:
On GTK, the RowActivated event of a ListBox is raised when the user performs a double-click on a row. To mimic this behavior on WPF, I used the MouseDoubleClick event of a System.Windows.Controls.ListBox.

References:

http://en.wikibooks.org/wiki/GTK%2B_By_Example/Tree_View/Events#Double-Clicks_on_a_Row
http://docs.go-mono.com/?link=E%3aGtk.TreeView.RowActivated
https://msdn.microsoft.com/en-us/library/system.windows.controls.control.mousedoubleclick(v=vs.110).aspx

### DIFF
--- a/Xwt.WPF/Xwt.WPFBackend/ListBoxBackend.cs
+++ b/Xwt.WPF/Xwt.WPFBackend/ListBoxBackend.cs
@@ -30,6 +30,7 @@ using System.Windows;
 using System.Windows.Controls;
 using Xwt.Backends;
 using Xwt.WPFBackend.Utilities;
+using System.Windows.Input;
 
 namespace Xwt.WPFBackend
 {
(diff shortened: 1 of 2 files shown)


sevoku:
I thought about exactly this way to implement this, but wanted to take a closer look and think about implementing this on the CellView side (handle cellviews clicks and notify listview).

Can this solution interfere with something, like multiselection? Can ```ListView.SelectedIndex``` point to an other row under some rare conditions?

(closes #286)

sevoku:
I've tested this again and it seems to be ok to use `SelectedIndex` in this case. The `TreeView` is more tricky, I've added it here: https://github.com/mono/xwt/commit/baedc7a472e92c6ba7869fb5fba87fcece8a4c2c.

Thanks!

PR #492, opened 2015-02-21
Appveyor integration

sevoku:
I have finally managed to automate the build process and NuGet packaging on AppVeyor. This PR contains various build and test (from #489) fixes and all things required for NuGet publishing.

What and how:
* AppVeyor hooks into GitHub and builds every new commit (@slluis: a webhook must be added)
* Automatic builds are enabled only for branches containing an appveyor.yml (can be changed, if required)
* Automatic builds are enabled for pull requests without NuGet publishing
* The automatic build succeeds (by pushing the artifacts) only if Xwt builds successfully. This includes:
  * Xwt
  * Xwt.Gtk
  * Xwt.Gtk3
  * Xwt.Wpf
* The tests (WpfTestRunner, GtkTestRunner) are run and the results are displayed in the build summary, but they must not pass for a successful AppVeyor build. This is because we still have many non-working tests, which have to be fixed (we can require passing, when the time comes).
* The dependencies (required NuGet nd Gtk packages) are cached for faster builds.

Versioning and NuGet:
* AppVeyor produces X.X.{build}-{branch} version numbers and patches all AssemblyInfo.cs files with the current build version
* Nuspec files take the AssemblyInformationalVersion (inserted by AppVeyor)
* This PR proposes a 1.0.X-master version with a prerelease NuGet tag for the master branch. Stable branches like an upcoming 0.2 should include own appveyor.yml with the according version at the top of the file (e.g. ```version: 0.2.0.{build}```) and without the ```{branch}``` variable in ```assembly_informational_version: '{version}'```.

NuGet specials:
The Xwt.Gtk*.dlls need their configurations. But simply including them in the same NuGet package folder does not work (the *.config file is not copied when the package is installed). For this to work correctly I have added some prop files (see http://docs.nuget.org/create/creating-and-publishing-a-package#user-content-import-msbuild-targets-and-props-files-into-project for more information)


sevoku:
Perhaps we should not use 1.0.X for master builds, but something like 0.9, or we increment the stable version by 0.+1. But using 0.X.{build} limits us a little bit in the patch version number, since NuGet does not like thinks like X.X.X.X (see https://docs.nuget.org/create/versioning and http://semver.org/, NuGet tpits warnings when packing and may have other issues on the consumer side). So it would be impossible to have patch versions like 0.2.2, cause the third one must be reserved for the AppVeyor build number.
Alternative: keep proposed 1.0.{build}-master for the current development and increment minor when we have stable 1.0.

slluis:
master should now be 0.2.x. Once we are ready for the 0.2 release, we create a branch for 0.2 and we bump master to 0.3, and so on.

PR #498, opened 2015-02-27
Fix default ColorSelectorBackend without window

sevoku:
This fixes the problem when the default ColorSelector is rendered without a parent window (e.g. when initialized with Wpf, or in a foreign host context). In this case the color box is drawn without taking the current screen scale factor into account.


sevoku:
I'm not sure about possible regressions. There could be problems on Mac in guest mode and with retina, maybe.

PR #500, opened 2015-02-28
[MAC] Respect SelectionMode on row selection

sevoku:
The default behavior when selecting single rows should be extending the selection when SelectionMode is set to ```SelectionMode.Multiple``` (WPF and GTK work this way). The current Mac behavior is to replace the selection (deselect previously selected rows), which is fixed here.


slluis:
We should have unit tests for this.

PR #505, opened 2015-03-12
[Mac] SpinButton Fixes

sevoku:
This fixes following problems with the spinButtonBackend on Mac:
* sizing / alignment
* Digits handling/formatting
* initial value
* handle text input/changes
* spin on key presses (up/down)

and adds following missing features:
* SpinButton.ValueChanged event
* Indeterminate support

And here are some Screenshots:
SpinButton Sample before/after:
SpinButton Sample before
SpinButton Sample fixed

And the fixed sizing/alignment issue (ListView sample before/after):
SpinButton sizing in ListView sample
SpinButton sizing in ListView sample fixed


slluis:
Looks good.

PR #506, opened 2015-03-14
Fix RadioButtonMenuItem group handling

### LINKED ISSUE #256: MenuItem is not working
Unfortunately i've found Xwt.MenuItem does not raise events properly. Furthermore, RadioButtonMenuItem doesn't work properly. I've found that because of  `enabledEvents` property in backend classes is null. This hack helps to solve this issue:
<code>
public void Initialize (IMenuItemEventSink eventSink)
{
            this.eventSink = eventSink;
            EnableEvent(MenuItemEvent.Clicked);
}
</code>

sevoku:
The group handling of RadioButton menu items relies on the Clicked event, which is disabled by default. This patch enables the Clicked event even if it is not subscribed. An extended Menu example is included, too.

@slluis: is this the right way to do it?

(closes #256)


SirCmpwn:
:+1:

PR #511, opened 2015-04-07
Fix build with Xamarin.Mac 2.0

mhutch:
* NSAttributedString constants have moved
* `System.Drawing.*F` to `CoreGraphics.CG*` conversions are gone

This will break the build with older versions. The conversions are fine but renamed constants won't work.

### DIFF
--- a/Xwt.XamMac/Xwt.Mac.CellViews/CanvasTableCell.cs
+++ b/Xwt.XamMac/Xwt.Mac.CellViews/CanvasTableCell.cs
@@ -76,7 +76,7 @@ namespace Xwt.Mac
 			var size = new CGSize ();
 			Frontend.ApplicationContext.InvokeUserCode (delegate {
 				var s = Frontend.GetRequiredSize ();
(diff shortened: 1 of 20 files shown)


slluis:
Will this still build for Xamarin.Mac < 2.0 ?

Therzok:
Nope. NSStringAttributeKey is a breaking change.

jstedfast:
Please please please can we merge this? I'm having to constantly switch between Xamarin.Mac 1.2 and 2.0 and this is driving me insane.

jstedfast:
I *only* need Xamarin.Mac 1.2 to build XS...

sevoku:
I can't test anything at the moment.

@slluis: maybe it is a good point to freeze a [0.2 branch](https://github.com/mono/xwt/milestones/v0.2) (or tag it at least) and proceed with Xamarin.Mac 2.0 on master, since there are some other breaking changes to check/merge? Should we proceed without CI and NuGet for Mac for now?

PR #520, opened 2015-06-14
Add TextInput to KeyEvent (was: Fixed WPF KeyCode)

cra0zy:
"(int)e.Key" just returned the enum value instead of ascii code

### DIFF
--- a/Xwt.Gtk/Xwt.GtkBackend/WidgetBackend.cs
+++ b/Xwt.Gtk/Xwt.GtkBackend/WidgetBackend.cs
@@ -638,7 +638,7 @@ namespace Xwt.GtkBackend
 			if ((args.Event.State & Gdk.ModifierType.Mod1Mask) != 0)
 				m |= ModifierKeys.Alt;
 
-			return new KeyEventArgs (k, (int)args.Event.KeyValue, m, false, (long)args.Event.Time);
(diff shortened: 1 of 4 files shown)


sevoku:
KeyEventArgs should be created with the native key code as the second argument and not its unicode or ascii representation:

```c#
public KeyEventArgs (Key key, int nativeKeyCode, ModifierKeys modifiers, bool isRepeat, long timestamp)
```

With your change KeyEventArgs.NativeKeyCode would not be correct.

What we need here is an additional String property in KeyEventArgs which should contain the characters, if any. And you can use the TextInput event, too.

cra0zy:
> With your change KeyEventArgs.NativeKeyCode would not be correct.

Sorry, I misunderstood what the number should be doing.

> What we need here is an additional String property in KeyEventArgs which should contain the characters, if any. And you can use the TextInput event, too.

Ok, easy enough, will update this tomorrow morning.

sevoku:
I can't test this atm. How about Mac and Gtk?

cra0zy:
> I can't test this atm. How about Mac and Gtk?

With this I only update Gtk and Wpf.

Since I don't own a Mac I can't update the Mac side so I just added a new constructor method for KeyEventArgs so I don't break something.

Issue #516, opened 2015-05-22
Multi-touch events


Gerharddc:
Is there any way to receive touch events using XWT? I am referring specifically to the multi-touch events that can be generated by WPF and GTK. I guess maybe those generated by the OSX multi-touch trackpads could also be useful.

sevoku:
Xwt has no support for touch events. I was planning to make some research and to add some basic api, but I have no time for it atm. See #406, too.

Gerharddc:
Ok I might look into adding it myself if I have the time. I have already done some research though and this should be fairly easy on WPF for Windows. GTK# will be a problem though because the current version of Gtk+ that has been wrapped does not appear to have touch events. Someone will have to update Gtk# first. After that it should be easy. On OSX there is off course no such thing as a touch screen but if I have read it correctly then Gtk+ can/might/will maybe recognize gestures from the multi-touch trackpads. But again this will only be after an update on that side.

sevoku:
I have no ea about Wpf, but Gtk seems to be possible. The major problem is Gtk#, you're right. There seems to be a PR (mono/gtk-sharp#128) for 3.12, but I have no idea if it works and contains the touch event + structures. And the gtk-sharp repo has not been updated since jan 2014. The other problem is, that if gtk-sharp gets the right apis, it would take pretty long time until the packages (linux/windows/mac) get updated. So I was thinking about mapping the events using GLib and adding the structures directly to Xwt.Gtk3. With some version checks it should be pretty easy with the current gtk-sharp without the apis, but I had no time for more work on this.

sevoku:
An I forgot: the main challange is to create some unified Touch-API in core Xwt. I don't know how different the native implementations are.

Gerharddc:
@sevoku I'll check out glib if I ever get time to try and work on this myself. What I can tell you from experience though is that a basic touch API that offers just a bunch of touch points with movement deltas and coordinates should be implementable for all platforms and be very helpful. All that is then needed is a touchstart, touchchange, and touchend event on a surface with these points passed as an array parameter. This is one way that I believe Qt, which run ons almost any platform, offers touch.

lytico:
I plan an implemenation based on this:
http://www.w3.org/TR/pointerevents

lytico:
first proposal: https://github.com/mono/xwt/pull/523

cra0zy:
> There seems to be a PR (mono/gtk-sharp#128) for 3.12

[...]

Issue #406, opened 2014-11-14
Xwt for Android


lytico:
I have started an android-backend:
https://github.com/lytico/xwt-mobile

Maybe some are interested to contribute.

sevoku:
Wouldn't it make more sense to go with Xamarin (and Xamarin.Forms), or is MonoDroid still alive?

DavidKarlas:
Imho... even desktop UIs are so different that is hard to make Xwt work nicely... Adding mobile into mix sounds like recipe for disaster... Maybe I'm wrong... Eto.Forms are working on this... But imo... Using PCL for business logic/models and 2 more libraries on top of that one with Xwt for desktop and Xamarin.Forms for mobile makes much more sense...

sevoku:
Yes, even having desktop and mobile support in one toolkit, requires different implementations (I've seen it in Eto and its far from being ready). The only interesting thing would be to be able to use Xwt widgets with Xamarin.Forms. Imagine you have developed some kind of custom widget and you could reuse the code on a mobile platform :+1:

DavidKarlas:
I really don't see much connection between this 2 UIs... on desktop you have at least 10 controls on screen with 3-5 in same row... On mobile you have max 10 controls and max 2 in same row... It's totally different UI. I can't even imagine scenario other then Login screen that makes sense...

sevoku:
The first and best example is drawing. It must not be a complete control, or even a form (you're right it really doesn't make sense in most cases). But think of things like plotting graphs and such.

lytico:
Indeed, I use it for the drawing part of my application. 
Currently I don't plan to implement the widget part of xwt. 
I don't think that the same xwt app could run on desktop and mobile, but most of the controls are very similar in both ui's.
One part missing in xwt is touch support, too. But implementation of this could be usefull on desktop too.

sevoku:
Touch support is something I'll need in near future. I've looked at Wpf and Gtk APIs and they are completely different. And Gtk# does not include the Gestures classes, so a custom native wrapper is required. No idea about Mac right now. The first step would be to define some Xwt API for touch handling, which covers all possible toolkit patterns.

Back to topic: I'm not really sure right now, if it could really make sense to implement the complete widget part (Sure, GUIs are completely different, but having one Toolkit for all of them could still be nice). Its more a philosophical question, I think. If someone makes a complete Droid or Xamarin.Forms backend and requests a pull, why not?.
[...]

PR #523, opened 2015-06-30
[Xwt] Touch-Support Api

lytico:
this is a proposal for touch event support in Xwt.
(see: https://github.com/mono/xwt/issues/516 and discussions in https://github.com/mono/xwt/issues/406)

This proposal is based on http://www.w3.org/TR/pointerevents

I want to discuss the api before starting the first implementation.

### DIFF
--- a/Xwt/Xwt.Backends/IWidgetBackend.cs
+++ b/Xwt/Xwt.Backends/IWidgetBackend.cs
@@ -341,6 +341,16 @@ namespace Xwt.Backends
 		/// <param name="args">The mouse scrolled arguments.</param>
 		void OnMouseScrolled(MouseScrolledEventArgs args);
 
+		void OnPointerEntered ();
+
+		void OnPointerExited ();
+
+		void OnPointerDown (PointerEventArgs args);
+
+		void OnPointerUp (PointerEventArgs args);
+
+		void OnPointerMoved (PointerEventArgs args);
+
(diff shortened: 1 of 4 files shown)


sevoku:
What is the exact difference to the mouse events? Only the additional event argument properties? Wouldn't it be easier to add the additional information to the existing mouse events?

As of my experience with Gtk, it produces same mouse events for all input methods (mouse/touch/pen). I think Wpf should do it the same way, too (some sort of mouse emulation for apps without touch support). So adding additional information about the pointer to the existing mouse events would be more practical I think. Tracking multiple pointers at the same time should be no problem with the additional PointerId, but it should be added to the Entered/Exited events, too. But with this pattern it keeps to be some sort of "mouse emulation" and it is up to the user to handle gesture recognition.

The really interesting thing in Gtk3 are Gestures (https://developer.gnome.org/gtk3/stable/GtkGesture.html) which can be used to handle real touch things touch-move, zoom, rotate, etc. Does Wpf (or some 3rd party lib) have something similar?

So I think the "pointer" Api should be kept simple (extend existing mouse api), the Multitouch/Gesture implementation must be handled by the user then. In a second step we could add some sort of Gesture API for a real Multitouch "experience", which would not depend on the "low-level" pointer/mouse events.

residuum:
Shouldn't this be closed?

Issue #514, opened 2015-05-16
Build break: NSStringAttributeKey does not exist


skrysmanski:
Commit ca15fd675b057bd536e2c766be4e4ff929eaa4ff broke Xwt.Mac (i.e. not XamMac) especially because `NSAttributedString` has been replaced at a lot of places with `NSStringAttributeKey` which the compiler can't find. Also `nfloat` definitions are missing in some files.

sevoku:
Yes, it's really a breaking change!

xajler:
It is actually not a breaking change, you need to have Xamarin.Mac 2 installed to build this one or any Mac related repository for mono, you can download it here https://forums.xamarin.com/discussion/42674/stable-service-release-xamarin-mac-2-0-1-bug-fixes-for-2-0-0#latest

sevoku:
Well its a breaking change for MonoMac users with the Xwt.Mac backend. We need some workaround here (maybe an extension method like its done with Gtk2/3 differences). Or MonoMac support could be dropped completely, if it's not under active development anymore (the last update was mono/monomac@36508cffdd4ce29f6c6e87f7c30d621d3cc512f3 on 19 Feb 15, so if XamarinMac is its official successor then this was the last OSS version of MonoMac?)

skrysmanski:
@xajler I don't have Xamarin.Mac installed (nor do I want to start the trial just to find out), so could you tell me in which assembly `NSStringAttributeKey` is defined in? Xamarin.Mac? If so, we need a workaround because Xamarin.Mac isn't referenced (nor should it IMHO) by Xwt.Mac.

@sevoku I hope MonoMac support isn't dropped as this would be bad for open source developers.

xajler:
@skrysmanski I think is all due to changes in Xamarin Studio and driven by Xamarin. To be honest I have no clue what is Xwt.Mac is, all I wanted is to compile MonoDevelop for which I haven't find any binaries, it seems is now only Xamarin Studio. 
MonoDevelop and Xamarin Studio seems have dependency on Xwt.Mac, it was breaking on this ``NSStringAttributeKey``, but I could compile when I've installed Xamrin.Mac. 
For what I've know you don't need to have any trial for Xamarin.Mac, because I don't have any subscription with Xamarin, and I had no problem to compile MonoDevolop and Xwt.Mac when I've installed this Xamarin Mac.

sevoku:
[...]

PR #534, opened 2015-07-19
Fixed MonoMac compile break

skrysmanski:
Fixed for build break for Xwt.Mac (#514).

### DIFF
--- a/Xwt.Mac/NSStringAttributeKey.cs
+++ b/Xwt.Mac/NSStringAttributeKey.cs
@@ -1,0 +1,43 @@
+﻿using MonoMac.Foundation;
+
+namespace Xwt.Mac
+{
+	public static class NSStringAttributeKey
+	{
+		public static NSString ForegroundColor
+		{
+			get { return NSAttributedString.ForegroundColorAttributeName; }
+		}
+
+		public static NSString UnderlineStyle
+		{
+			get { return NSAttributedString.UnderlineStyleAttributeName; }
+		}
+
+		public static NSString BackgroundColor
+		{
+			get { return NSAttributedString.BackgroundColorAttributeName; }
+		}
+
+		public static NSString Obliqueness
+		{
+			get { return NSAttributedString.ObliquenessAttributeName; }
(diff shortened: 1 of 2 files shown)


Xwt.Mac/NSStringAttributeKey.cs
```diff
@@ -0,0 +1,43 @@
+﻿using MonoMac.Foundation;
+
+namespace Xwt.Mac
+{
+	public static class NSStringAttributeKey
+	{
+		public static NSString ForegroundColor
+		{
+			get { return NSAttributedString.ForegroundColorAttributeName; }
```
Therzok: This will definitely break Xamarin.Mac, since this was removed.

We need to introduce some define constants and most likely a new configuration to handle MonoMac and provide extension/wrapper API inside Xwt that provides compatibility.

skrysmanski: @Therzok  Could you elaborate on how exactly this would break Xamarin.Mac? Because the code is only in Xwt.Mac, not in Xwt.XamMac.

Therzok: Ah, sorry. I checked now and there are multiple configurations.

sevoku:
Works great, but I would propose 2 more things:
1. move the NSStringAttributeKey.cs to the Xwt.Mac subdirectory (project and filesystem: Xwt.Mac/Xwt.Mac/NSStringAttributeKey.cs)
2. add all NSStringAttributeKey members, so we have a complete conversion for the future

sevoku:
I've added the changes in https://github.com/skrysmanski/xwt/compare/fix-monomac-buildbreak...sevoku:fix-monomac-build, feel free to test and pull them into this PR.

@Therzok: could you take a look at 483e656044d13399f1b021d18f123ddec82dfbc2, not all members are supported with MonoMac, so I have imported the generated ones from Xamarin.Mac including the Field attributes.

sevoku:
Thanks!

PR #538, opened 2015-07-20
Improved detection of primary screen by using the appropriate property.

skrysmanski:
May be a little bit more robust than using reference comparison.


residuum:
This could also be merged.

residuum:
This could also be merged.

sevoku:
LGTM, Thanks!

Issue #535, opened 2015-07-19
Initial window location not centered on HiDPI displays on Windows


skrysmanski:
Currently, when having a HiDPI display on Windows and using Xwt.Wpf, the initial window position is not centered but rather in the lower right corner.

The reason for this is that `Desktop.PrimaryScreen.VisibleBounds.Center` on Windows returns coordinates in physical pixels (device-dependent pixels), while WPF (via `window.Left` and `window.Top`) expected logical pixels (device-independent pixels). I checked on OS X, the Xwt.Mac backend returns the screen size in device-independent pixels.

The documentation of `DesktopBackend` says that the methods should return sizes in "desktop coordinates". From just looking at the documentation, it's not very clear whether this means device-dependent or device-independent pixels - or just "the pixels type used by the backend"; maybe this should be better documented.

As for WPF backend, I'd suggest that `WpfDesktopBackend` is changed to device-independent pixels as expected by WPF. Scaling the screen *size* should be easy. However, I don't know how this should affect the screen's *location* - especially when there are multiple screens with different DPI scalings. Maybe the screen's location could be removed altogether from the screen API?

sevoku:
I think the dimensions should be unified at least on the backend level. So yes, WPF DesktopBackend should use device-independent pixels like the others.

Concerning the Screen API: there is a ScaleFactor property, which can be used for conversion. It should support multiple screens with different dpi settings, too. But I have no HiDPI screen to test.

skrysmanski:
Screen API on Windows seems not so easy to fix. I've created a separate issue (#539) for this.

PR #542, opened 2015-07-25
Fixed WPF window centering on HiDPI screens

skrysmanski:
Fix for #535


sevoku:
First time I was able to test with HiDPI. Looks good and all window tests pass with and without HiDPI.
Thanks!

PR #544, opened 2015-08-30
Optimize Font.FromName parser

sevoku:
For some fonts a valid registered name can include the font settings (weight, size, etc.). In such cases the parser fails to retrieve the font from the backend by stripping the settings from the name.

Fix: the name parser should try to validate the full name including provided settings first.

Examples:
* "Arial Rounded MT Bold", "Bold" is part of the font name and the registered weight is normal (there is only a "Regular" font face registered)
* "Bodoni 72": 72 is not the size but part of the name

### DIFF
--- a/Xwt/Xwt.Drawing/Font.cs
+++ b/Xwt/Xwt.Drawing/Font.cs
@@ -121,6 +121,8 @@ namespace Xwt.Drawing
 
 		internal static Font FromName (string name, Toolkit toolkit)
 		{
+			if (string.IsNullOrWhiteSpace (name))
+				throw new ArgumentNullException (nameof (name), "Font name cannot be null or empty");
 			var handler = toolkit.FontBackendHandler;
 
 			double size = -1;
--- a/Xwt/Xwt.Drawing/Font.cs
+++ b/Xwt/Xwt.Drawing/Font.cs
@@ -181,13 +181,8 @@ namespace Xwt.Drawing
 			string[] names = fontNames.Split (new [] {','}, StringSplitOptions.RemoveEmptyEntries);
 			if (names.Length == 0)
 				throw new ArgumentException ("Font family name not provided");
-
-			foreach (var name in names) {
-				var n = name.Trim ();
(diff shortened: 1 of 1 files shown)


Xwt/Xwt.Drawing/Font.cs
```diff
@@ -131,13 +131,18 @@ internal static Font FromName (string name, Toolkit toolkit)
 			int i = name.LastIndexOf (' ');
 			int lasti = name.Length;
 			do {
+				if (lasti > 0 && IsFontSupported (name.Substring (0, lasti)))
+					break;
+				
 				string token = name.Substring (i + 1, lasti - i - 1);
```
Therzok: I know this probably isn't related to the PR, but what if the name is empty?

We'd end up with `name.Substring (1, -1);` on a 0 Length string. Not ok. We should probably assert a `string.IsNullOrWhitespace (name)` at the beginning.

Xwt/Xwt.Drawing/Font.cs
```diff
@@ -167,6 +172,22 @@ internal static Font FromName (string name, Toolkit toolkit)
...
 
+		static bool IsFontSupported (string fontNames)
+		{
+			LoadInstalledFonts ();
+
+			string[] names = fontNames.Split (new [] {','}, StringSplitOptions.RemoveEmptyEntries);
+			if (names.Length == 0)
+				throw new ArgumentException ("Font family name not provided");
+
+			foreach (var name in names) {
```
Therzok: Style: `names.Any (name => installedFonts.ContainsKey (name.Trim ()));`

Therzok:
Looks good other than my 2 comments.

slluis:
Is it possible to add unit tests for this?

sevoku:
The problem would be that for useful unit tests the testing system must have some problematic fonts installed, but only Mac has preinstalled fonts with such names. We could add platform specific tests for ```IsFontSupported```, include some problematic font names (by registering them with using ```RegisterFontFromFile```) and test it for all installed fonts.

PR #546, opened 2015-09-08
Code formatting

steffenWi:
Correctly formatted all .cs files in the Xwt.WPF, WpfTest and XWT projects.


sevoku:
This is a good idea! But it should consist of several different commits with appropriate changes (one commit should include the project configuration for new files only and then one commit for each project with the formatting changes, so everything can be reproduced easily).

steffenWi:
> (one commit should include the project configuration for new files only and then one commit for each project with the formatting changes, so everything can be reproduced easily

I don't understand that part. I'm not adding new files here, neither am I changing the project configuration.

sevoku:
Well, I didn't go through all changes. I assumed that you could have changed the formatting settings of the projects (MD project settings dialog). If not, you should double check if they fit your changes, so we have a consistent configuration here for new files.

BUUUT: I don't think that we can merge this that easily anyway. It will break most of the PRs and make it necessary to rebase all wip branches too. So I think it is a very good idea, but this (reformatting, not merging this PR!) has to be done by one of the contributors while merging one of the larger changesets (maybe together with the image-loader or dark-skin branches, when they're finished, @slluis).

PR #420, opened 2014-12-16
Iconify/minimize windows

rpc-scandinavia:
I discovered that it only is possible to maximize the window to full screen.
I implemented the "Ifonify" method and the "WindowState" property.
I am not able to test the Mac code, but it compiles on my Linux with MonoDevelop and a copy of "MonoMac.dll".

Issue #547, opened 2015-09-12
Programatically maximize a window


AngeloG:
Is this possible? There's a `Fullscreen` property on `Window` but it doesn't seem to be maximizing.

sevoku:
It should work. Which backend are you using?

AngeloG:
It maximizes the window yes, but (on WIndows/WPF, at least) it also overlays the taskbar and hides the maximize/minimize buttons.

sevoku:
Ah, ok. You're right, see: #420. It is not implemented yet.

sevoku:
Could you try whether #420 works for you? I think it needed only a rebase, then I could merge it next days.

AngeloG:
It seems like `Xwt/Xwt/WindowState.cs` is not added to the project, so it is giving errors on compiling. Other than that, `FullScreen = true;` and `WindowState = WindowState.FullScreen;` works as expected on Windows/WPF.

sevoku:
Thanks for testing. I'll take a look.

sevoku:
Ok, it was a little bit more complex. I've improved #420 in #549. The problem with Wpf is that it does not support FullScreen and Maximized at the same time. Which one is performed depends on the window type. Decorated windows always maximize (taskbar visible) and undecorated windows are always fullscreened (no taskbar). I've added a little workaround for normal windows, but borderless windows support FullScreen only (Maximized becomes always FullScreen).

PR #549, opened 2015-09-13
Finish WindowState implementation (was #420)

sevoku:
This is a complete version of the WindowState implementation (closes #420) including:
* renamed Window.Iconify to Window.Iconified
* renamed WindowState.Icon to WindowState.Iconified
* new WindowState.Maximized
* completed backend implementations
* rebased

Limitations/Workarounds:
* Wpf: it depends on the window style (decorated) whether a window can be maximized or fullscreened. A workaround for normal windows is included (removes/restores decorations when FullScreen is toggled)
* Mac: When switching from FullScreen to Maximized or Iconified, the new state has to be set after the Unfullscreen action is finished. The workaround connects the ```DidExitFullScreen``` event to delay the state change.

### DIFF
--- a/TestApps/Samples/Samples/Windows.cs
+++ b/TestApps/Samples/Samples/Windows.cs
@@ -37,9 +37,24 @@ namespace Samples
 			bp.Clicked += delegate {
 				Window w = new Window ();
 				w.Decorated = false;
-				Button c = new Button ("This is a window");
+				Button c = new Button ("Close");
 //				c.Margin.SetAll (10);
-				w.Content = c;
+
+				VBox box = new VBox();
+				var biconifiy = new Button ("Iconify");
+				var bmaximize = new Button ("Maximize");
+				var bfullscreen = new Button ("Fullscreen");
+
+				biconifiy.Clicked += (sender, e) => w.Iconified = !w.Iconified;
+				bmaximize.Clicked += (sender, e) => w.Maximized = !w.Maximized;
+				bfullscreen.Clicked += (sender, e) => w.FullScreen = !w.FullScreen;
+
+				box.PackStart (biconifiy);
+				box.PackStart (bmaximize);
+				box.PackStart (bfullscreen);
+				box.PackStart (c);
+
+				w.Content = box;
 				c.Clicked += delegate {
 					w.Dispose ();
 				};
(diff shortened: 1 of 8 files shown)


sevoku:
(closes #547)

elsupergomez:
"Show borderless window" work fine. But I want to set the state of the window when create the window before shown, like this:

var w = new Window {WindowState = WindowState.Maximized };

and this break. Since WindowState is a property, should be possible assigned the value before shown the window.

sevoku:
Ok, maybe we can delay the assignment using the Realized handler...

sevoku:
I have optimized the state handling. Setting the state before the window is shown should work and It is more reliable on Mac now. @elsupergomez could you verify?

elsupergomez:
Probe far, this seems to work well.
@sevoku it is verified by me.

slluis:
This API looks a bit weird to me. Why do we need a PreviousWindowState property?
I don't think we need properties for Iconified and Maximized if we have a WindowState property.

elsupergomez:
My opinion. PreviousWindowState is necessary when want restore window state from Iconify. The user could use a normal, maximized or fullscreen windows, and then Iconify. We need to know how to restore the non Iconify previews state of windows when deiconify. The same thing happen for fullscreen to other state.

About properties for Iconified and Maximized, I think same slluis. That could be handle with WindowsState property

sevoku:
WRT `Iconified`, `Maximized` and `FullScreen`: I agree, but on the other hand its sometimes easier to set unset one of them. However, initially I decided to keep them (not break the API) and added `PreviousWindowState` to track them (setting unsetting one the bools requires to know the previous state). Since the backends handle it correctly now, we can kill these properties.

WRT `PreviousWindowState`: Its exactly for what @elsupergomez described, but its not that simple to implement, I have been thinking about to implement a `WindowStateChanged (previousState)` event, BUT the problem is that its really hard to track state changes using events. Not all transitions are signalled by the toolkits (for example Gtk does not signal the Maximized state, and some other states depend on the WM). The current implementation tracks at least transitions done programmatically and is more reliable than event handling, but its still not 100% reliable for changes done by the WM (when the user clicks one of the titlebar buttons). Therefore the example includes a `BoundsChanged`handler which can be used to verify the states.

Therzok:
@sevoku ConfigureEvent should have https://developer.gnome.org/gdk3/stable/gdk3-Event-Structures.html#GdkWindowState

sevoku:
@Therzok yes, it has (afaik its the same as handling `WindowStateEvent` cause Gtk passes state changes directly through it, at least this is the experience I made), but it still depends on the WM and the last time I tried to handle it on Mac, it didn't work as expected. Even on Linux its not guarateed to be reliable.

PR #551, opened 2015-09-24
[Xwt.XamMac] Add application dock menu support to AppDelegate

iainx:
Adds an event to the AppDelegate to return custom dock menu items

### DIFF
--- a/Xwt.XamMac/Xwt.Mac/MacEngine.cs
+++ b/Xwt.XamMac/Xwt.Mac/MacEngine.cs
@@ -271,6 +271,7 @@ namespace Xwt.Mac
 		public event EventHandler Unhidden;
 		public event EventHandler<OpenFilesEventArgs> OpenFilesRequest;
 		public event EventHandler<OpenUrlEventArgs> OpenUrl;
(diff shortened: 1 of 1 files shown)


Xwt.XamMac/Xwt.Mac/MacEngine.cs
```diff
@@ -376,4 +390,9 @@ public OpenUrlEventArgs (string url)
 			Url = url;
 		}
 	}
+
+	public class ShowDockMenuArgs : EventArgs
+	{
+		public NSMenu dockMenu;
```
slluis: This should be a property, and first letter should be upper case

iainx: Yeah, not sure what I was thinking here. Fixed it in the new push

sevoku:
@iainx whats the status here? Could it be still useful, even if not used in MD?

Therzok:
Ping on this, would be good to merge so we can take a look at:
https://github.com/mono/monodevelop/pull/1065

sevoku:
LGTM, lets merge. This could be also useful for a XWT wide application menu, Gtk has also support for application menus. We'd need an api for a main app window...

PR #560, opened 2015-11-03
Add Sorting functionality to WPF ListView

steffenWi:
- ListView.cs: Added method with functionality to satisfy interface change and add functionality so that sorting a ListView by clicking the column's header will work
- ITableViewBackend: adjusted interface to support sorting
- ListDataSource.cs and ListViewBackend: Added functionality so that the data source/the listview can be sorted. LINQ is used to actually order/sort the data.


steffenWi:
Had to fix a doubled constructor that snuck in during cherry-picking the commit.

sevoku:
I get an NRE in the ListView example.

PR #561, opened 2015-11-03
Wpf combobox with free textentry

steffenWi:
With this change it's possible to enter a free text into a WPF combobox. The text can then, by the application using XWT, be added as a new item to the combobox, for example.


steffenWi:
Anyone?

sevoku:
What is the benefit of using this compared to a `ComboBoxEntry` which has much more features for text editing?

steffenWi:
I wasn't aware I could do that by using ComboBoxEntry. I've not used that class and I'm not aware of it.

sevoku:
Well, the functionality is given with `ComboBoxEntry`, take a look at the samples in https://github.com/mono/xwt/blob/master/TestApps/Samples/Samples/ComboBoxes.cs#L82. If you experience some limitations there (compared to this patch) feel free to patch/PR for ComboBoxEntry. I don't think that we'll add any editing features to the simple `ComboBox`.

PR #562, opened 2015-11-03
[WPF]Set the name of a control

steffenWi:
* Implemented extended interface in WPF/WidgetBackend.

This change is important for UI automation/UI testing tools like Microsoft Test Manager. These tools use the name of a control to identify it, without setting a name, at least Microsoft Test Manager will use the first control of that type that it encounters (for example: the minimize button)

### DIFF
--- a/Xwt.WPF/Xwt.WPFBackend/WidgetBackend.cs
+++ b/Xwt.WPF/Xwt.WPFBackend/WidgetBackend.cs
@@ -178,6 +178,12 @@ namespace Xwt.WPFBackend
 			set { Widget.Opacity = value; }
 		}
 
+		public string Name
+		{
+			get { return Widget.Name; }
+			set { Widget.Name = value; }
+		}
+
 		FontData GetWidgetFont ()
 		{
 			if (!(Widget is Control)) {
--- a/Xwt.WPF/Xwt.WPFBackend/WindowFrameBackend.cs
+++ b/Xwt.WPF/Xwt.WPFBackend/WindowFrameBackend.cs
@@ -59,6 +59,7 @@ namespace Xwt.WPFBackend
 			Initialize ();
 		}
 
+		public object Tag { get { return window.Tag; } set { window.Tag = value; } }
 		public ApplicationContext Context { get; private set; }
 
 		public virtual void Initialize ()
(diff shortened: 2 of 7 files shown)


steffenWi:
Anyone?

sevoku:
Great idea, but:

1. Why override ```XwtComponent.Tag``` in 548fd036af61c66e4a234158bdea97ede547a54f? Consider using same pattern as for ```Widget``` by introducing a ```WindowFrameBackend.Name``` property (mapped to the ```window.Tag``` Wpf property).
2. Missing Gtk / Mac impl.

steffenWi:
For 1.: Because I didn't want to break existing functionality while I also didn't want to implement it for every control.

For 2.: true. Forgot that I adjusted an interface. That'll take some time...

sevoku:
For 1.: why would you break existing functionality by adding a `Name` property to WindowFrameBackend? (And you have implemented `Name` for all other controls already)

sevoku:
For 2: Gtk is pretty simple cause all Gtk widgets have a `Name` property. There is nothing comparable in MonoMac/Xamarin.Mac, so there the property could be simply stored inside the backend wrapper class without passing it to any native backend property (Just declare a ```public string Name {get; set; }``` in Widget backend).

steffenWi:
For 1: that was about the .Tag property. Not .Name or have I missed something?

sevoku:
For 1: you have introduced a new `Widget.Name` property in 2f45218, right? But you reuse the `XwtComponent.Tag` property for the same purpose in WindowFrame. For consistency reasons you should introduce a new `WindowsFrame.Name` property too, instead of reusing `Tag` *only* for WindowFrame.

sevoku:
Partially merged: Name property from 2f452188da84fc81d04154af88bd08c3540c3acc (via 4adb26fdc17fe172b5de753bb31563becd37c465)
WindowFrame.Name implemented in dff62066e272d6d9e01bc119a48ce8fe6650185c

I don't know whether the WindowFrame.Tag (btw all FrameworkElements have a Tag property so maybe not only Windows should implement that) property might be useful. But there is nothing comparable in Gk/Mac, so it would only work with Wpf wich wouldn't make much sense to implement at all.

PR #566, opened 2015-12-27
Fix getting images from clipboard in Gtk

pleonex:
This fixes `Clipboard.GetImage()` for the Gtk backend.

I have made visible the internal methods from the Xwt project for Xwt.Gtk. This is a requirement to access to the internal constructor of `Image` that accepts `GtkImage` objects (`clipboard.WaitForImage ()` returns a `Gtk.PixBuf` object). I am not sure if this is the best approach but it's working for me. I have copied the public key from a comment in *Xwt.Gtk/AssemblyInfo.cs*.


sevoku:
You can use ```ApplicationContext.Toolkit.WrapImage (new GtkImage (clipboard.WaitForImage()));``` without making the constructor public, I think.

pleonex:
That's right, I have fixed it. Thanks!

Therzok:
Lgtm!

Issue #526, opened 2015-07-02
Event TextChanged in TextCellView in ListView


ezequiel89:
When I use an editable TextCellView ,I put a new value, the event TextChanged have the old value. And the new value is going to appear the next time that I call/invoke the event.

I need that the event handler should have access to the new value on theTextCellVIew when it has raised the TextChanged event.

sevoku:
It looks like the new text is applied only when ITextCellViewFrontend.RaiseTextChanged returns true, which raises the event and returns its WidgetEventArgs.Handled value. Inside the event handler the old text is still inside the store. I think we need a TextChangedEventArgs parameter, which should hold the new text, so it can be used inside the event handler.

And it looks like this is only implemented with Gtk...

ezequiel89:
Thanks for your prompt reply.

PR #570, opened 2016-01-06
Fix Event TextChanged in TextCellView in ListView #526

juandinamarca:


residuum:
I see two problems with it:
- Adding `WidgetEventArgs.Text`. Why not use a different class, e.g. `TextWidgetEventArgs` with the properties `bool Handled` and `string Text`?
- This is only an implementation for GTK, not for WPF or Mac.

PR #571, opened 2016-01-10
Fix mistype in CustomCellRendererImage

m13253:
To fix compile error in target `Xwt.Gtk3`, `Xwt.Gtkbackend.ImageRenderer.OnRender`,
change `GetSize` to `OnGetSize`.

The compile error message was:
```
xwt/Xwt.Gtk/Xwt.GtkBackend.CellViews/CustomCellRendererImage.cs(4,4): Error CS0103: The name `GetSize' does not exist in the current context (CS0103) (Xwt.Gtk3)
```


sevoku:
Thanks! But it should be `Gtk3Workarounds.GetSize()` which invokes the internal width-for-height/height-for-width logic in Gtk3, see 11c4bb6287fa63c8433d57fb50f4772aebbd3d04. Hope it works as expected.

PR #572, opened 2016-01-10
Add new Menu.Font property

### LINKED ISSUE #569: Mac: Better small PopupContextMenu handling
Currently we have small `PopupContextMenu` on Mac, and it would be better to have an API to select the font size (I guess standard or small, there's no reason to allow any font size value).

Here's my old patch from last year: https://github.com/mono/xwt/commit/4e54e814eff07ca09646b387f16ee96d1dbfecc3

Thanks,
V.

sevoku:
The new `Menu.Font` property allows to modify the font in menus. The font will be applied on all items of a menu, but not on submenus, so each new `Menu` instance will start with the default system font.

Includes WPF/GTK/MAC implementations.
(fixes #569)

### DIFF
--- a/Xwt.Gtk/Xwt.GtkBackend/MenuBackend.cs
+++ b/Xwt.Gtk/Xwt.GtkBackend/MenuBackend.cs
@@ -70,6 +70,23 @@ namespace Xwt.GtkBackend
 				return bar;
 			}
 		}
+
+		Pango.FontDescription customFont;
+
+		public virtual object Font {
+			get {
+				return customFont ?? menu.Style.FontDescription;
+			}
+			set {
+				customFont = (Pango.FontDescription) value;
+				foreach (var item in menu.AllChildren) {
+					var bin = item as Gtk.Bin;
+					if (bin != null)
+						foreach (Gtk.Widget w in bin.Children)
+							w.ModifyFont (customFont);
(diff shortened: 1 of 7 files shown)


Xwt.Gtk/Xwt.GtkBackend/MenuBackend.cs
```diff
@@ -70,6 +70,23 @@ public void InitializeBackend (object frontend, ApplicationContext context)
...
+		Pango.FontDescription customFont;
+
+		public virtual object Font {
+			get {
+				return customFont ?? menu.Style.FontDescription;
+			}
+			set {
+				customFont = (Pango.FontDescription) value;
+				foreach (var item in menu.AllChildren) {
+					var bin = item as Gtk.Bin;
```
Therzok: Shouldn't this be `Gtk.Container`? Gtk.Bin is a specialized single-child container.

Therzok:
LGTM except that gtk container type.

Xwt.Gtk/Xwt.GtkBackend/MenuBackend.cs
sevoku: Right

sevoku:
Thanks @Therzok!

PR #574, opened 2016-01-22
[Window] Fix leak of whole dialog/window content.

Therzok:
Basically, this disposes of the Window content in the Xwt hierarchy.

Previously, this was insufficient, while it _did_ remove the Gtk
references, subscribing to an eventhandler and overriding Dispose on a
widget that's placed in a Xwt Window will cause the widget to not be
disposed, thus leak memory.


Therzok:
tl;dr Xwt Dialogs/Windows never called dispose on its contents.

PR #575, opened 2016-02-03
[Gtk] Fix FocusInEvent keeping Popover in memory after it is disposed.

mrward:
Showing a Popover window and then disposing it when it is closed would leave the Popover window in memory if its TransientFor window was not closed. This was because the event handler added to the TransientFor window's FocusInEvent was not removed when the Popover was disposed.


Therzok:
Lgtm!

PR #576, opened 2016-02-08
Fix font toolkit validation

sevoku:
This fixes two different bugs with Fonts:

1. when setting the font size, `WithSize` should use its own `ToolkitEngine` instead of `Toolkit.CurrentEngine`
2. When a system font is being validated and initialized for a different toolkit, only its family name should be checked whether its a system font or not and the actual settings should be copied from the source backend. Otherwise a new font with the original system font family was created, which is not necessarily the system font of the target toolkit and which may not be even available for it (i.e. San Francisco fonts on OSX 10.11, which is not supported by Pango).


slluis:
LGTM

PR #577, opened 2016-02-11
[Gtk] Always take the custom drawn image for Multisize

sevoku:
This is a quick workaround for the case when a multisize
image is created from a set of custom drawn images.
GtkImage does not support combinations of custom drawing
and bitmap frames for the same image. In that case the
custom drawn multisize image is selected.


Therzok:
Probably will solve the 3rd most reported exception in VSMac. cc @slluis

PR #578, opened 2016-02-14
[Mac] Support different row heights in Lists/Trees

sevoku:
Previousy the height was fixed to the height of the highest row, because NSTableView/NSOutlineView use always the fixed RowHeight property in the deprecated NSCell based mode.
Now we use protocol delegates to calculate the height for each row dynamically.

The difference can be verified in the FontSelector widget, where the rows have different heights due to different font metrics. The Font parsing bug (fixed in #544) makes it even worse by rendering some fonts with extremely large sizes.
Before:
bildschirmfoto 2016-02-14 um 12 26 38
After:
bildschirmfoto 2016-02-14 um 12 24 10


Therzok:
:ship: :shipit:

PR #581, opened 2016-02-27
Update WindowFrame.cs

OmidID:
If you update Width or Height property instead of Size, it's will be change other side size to minimum value
Ex: this.Width = 100; // Height will be 0 now!

### DIFF
--- a/Xwt/Xwt/WindowFrame.cs
+++ b/Xwt/Xwt/WindowFrame.cs
@@ -383,6 +383,13 @@ namespace Xwt
 
 		internal virtual void SetBackendSize (double width, double height)
 		{
+			//Fix problem of change Width or Height property
+			if (width == -1)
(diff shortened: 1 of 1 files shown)


sevoku:
This looks like a WPF specific bug to me and should be fixed in https://github.com/mono/xwt/blob/master/Xwt.WPF/Xwt.WPFBackend/WindowFrameBackend.cs#L234 (see how its handled in [Gtk](https://github.com/mono/xwt/blob/master/Xwt.Gtk/Xwt.GtkBackend/WindowFrameBackend.cs#L125) and [Mac](https://github.com/mono/xwt/blob/master/Xwt.XamMac/Xwt.Mac/WindowBackend.cs#L414))

OmidID:
Oh I see. you right.
I close it and make a request there

PR #582, opened 2016-02-28
Update WindowFrameBackend.cs

OmidID:
If you update Width or Height property instead of Size, it's will be change other side size to minimum value
Ex: this.Width = 100; // Height will be 0 now!

https://github.com/mono/xwt/pull/581


sevoku:
Cherrypicked in 65be4635e7f34efff91bbcc6f3498f2b493487df (fixed style, no format changes)
Thanks!

PR #585, opened 2016-03-07
[GtkBackend] Add custom Gtk.MissingIcon for Gtk3

hwthomas:
For Gtk3 (as well as for Gtk2), draws a custom Gtk.MissingIcon if Stock icon cannot be loaded


sevoku:
Thanks!

PR #586, opened 2016-03-11
[Gtk.Mac] Remove broken Popover window border

sevoku:
This PR removes the border around the Gtk Popover window on Mac.

Before:
Unpatched: wrong border

After:
Patched: no border

### DIFF
--- a/Xwt.Gtk.Mac/GtkMacPopoverBackend.cs
+++ b/Xwt.Gtk.Mac/GtkMacPopoverBackend.cs
@@ -1,10 +1,10 @@
 ﻿//
-// GtkMacEngine.cs
+// GtkMacPopoverBackend.cs
 //
 // Author:
-//       Lluis Sanchez Gual <lluis@xamarin.com>
+//       Vsevolod Kukol <sevo@sevo.org>
 //
-// Copyright (c) 2014 Xamarin, Inc (http://www.xamarin.com)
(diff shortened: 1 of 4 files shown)


vancura:
@sevoku: Could you please grab me a screenshot of larger portion of the screen? I can't judge if there's a shadow or not. Missing shadow could be a problem.

sevoku:
Thats the point, there is no shadow:
bildschirmfoto 2016-03-14 um 09 21 09
The problem is that the border seems to be part of Cocoas shadow drawing impl. It is not possible to disable the border only.

vancura:
It's not as bad as I expected :)

That yellow, I suppose it's a custom styling which doesn't stay everywhere, right? :)

Can we have a stroke?

sevoku:
Yes, yellow is a custom example. This is how it would look with the default color + 1px border:
bildschirmfoto 2016-03-14 um 09 35 08

vancura:
This is nearly perfect. Can you specify stroke color to `rgba(0, 0, 0, 0.2f)`? And it would be good to have stroke width 1pt (2px on retina) since that's what we have on other controls?

sevoku:
Like this?
bildschirmfoto 2016-03-14 um 09 46 55

vancura:
Perfect, thanks.

slluis:
LGTM

PR #587, opened 2016-03-14
[Gtk] Add Popover transparency without alpha support

sevoku:
Simulates alpha color support by rendering the Gdk window content behind a popover on its background, if the Gdk screen does not support alpha colors or window opacity.

This is especially relevant on Windows where Gtk has no opacity/alpha color support (and may be triggered on Linux or Mac in rare cases, where i.e. the graphic adapter support is broken or an old Gtk version is being used):

Without Fake Transparency:
xwt popover broken alpha

With this Patch:
xwt popover fake alpha

And the fixed Xamarin Android Designer on Windows:
xwt popover android designer

### DIFF
--- a/Xwt.Gtk/Xwt.GtkBackend/PopoverBackend.cs
+++ b/Xwt.Gtk/Xwt.GtkBackend/PopoverBackend.cs
@@ -127,16 +127,52 @@ namespace Xwt.GtkBackend
 
 			protected override bool OnDrawn (Context cr)
 			{
+				bool withRoot = true;
 				int w, h;
 				this.GdkWindow.GetSize (out w, out h);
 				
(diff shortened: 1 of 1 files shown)


sevoku:
I have tested this on all platforms with all resolutions now. On Mac and Windows we can't use `gdk_cairo_set_source_window()` at all and with HiDPI the alternative `Pixbuf` rendering does not work.

For any non-working case, I've added a fallback which simply fills the whole window:
Popover rendering fallback
(this is just a demo, on Mac we have full alpha support with all resolutions)
But actually the fallback triggers on HiDPI Windows only.

@lluis: I hope this is robust enough now, since I catch all cases that are known to break.

PR #588, opened 2016-03-15
Fix Gtk3 Build

cra0zy:
Fixes 2 things:
1) Build output for Gtk3 Release was set to "Debug" folder instead of "Release"
2) Gtk.Object does not exist in Gtk 3, GLib.Object does however exist in both Gtk 2 and 3


sevoku:
Thanks!

PR #593, opened 2016-04-15
[Designer.IOS] Makes less restrictive parse when the result backend i…

netonjm:
…s not the expected


alanmcgovern:
I said it in slack, but better to say it here: the bug we're hitting is that we have gtk backend objects when we're in the xammac code paths. We need to fix the calling code to not make this mistake. This fix is incorrect as we should never have gtk widgets when inside here anyway!

netonjm:
ok, thanks alan!

PR #599, opened 2016-05-24
Implement menu accelerators

cra0zy:
Only implements it for Gtk side of things, WPF and Mac only have stubs.

### DIFF
--- a/TestApps/Samples/MainWindow.cs
+++ b/TestApps/Samples/MainWindow.cs
@@ -39,9 +39,10 @@ namespace Samples
 			
 			var file = new MenuItem ("_File");
 			file.SubMenu = new Menu ();
-			file.SubMenu.Items.Add (new MenuItem ("_Open"));
-			file.SubMenu.Items.Add (new MenuItem ("_New"));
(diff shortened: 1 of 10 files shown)


Xwt/Xwt/Key.cs
```diff
@@ -27,6 +27,7 @@
 
 namespace Xwt
 {
+	[Flags]
```
Therzok: \Why was this changed to `Flags`?

cra0zy: So multiple `Key` can be sent as accelerator, I could have also made `Accelerator` class to take in an array of `Key`, but that is really memory inefficient compared to this.

cra0zy: Then again most shortcuts only use one key and muliple modifiers... I can change it if you want.

Therzok: If I do `a | b`, it'll result in a totally other key, not combine the two keys. I'm possibly missing the point here, but having `Flags` won't make keys combinable.

cra0zy: Good point, I never looked into `Key` values and I always assumed it was a flag in Gtk like modifier keys... sorry about that.

Therzok: No worries, just wanted to know what was up. Glad we got it solved!

PR #603, opened 2016-06-29
[GTK] Fix Uri drop on Windows

Libco:
When you drop files on Windows the last splinted string is always "\0".


sevoku:
I've fixed it in 2d4a5db. The real problem here was that Data can contain anything, not only strings/uris. The right approach is to use `gtk_selection_data_get_uris`, but the managed gtk-sharp part is broken (it uses a string instead of a null term array).

PR #605, opened 2016-06-30
Rectangle.Contains(point) and Rectangle.Contains(rect) are inconsistent

lytico:
Contains (Rectangle rect)
includes the whole rect 

Contains (double x, double y)
excludes right and bottom borders of rectangle

this is inconsistent

### DIFF
--- a/Xwt/Xwt/Rectangle.cs
+++ b/Xwt/Xwt/Rectangle.cs
@@ -104,8 +104,8 @@ namespace Xwt
 		
 		public bool Contains (double x, double y)
 		{
-			return ((x >= Left) && (x < Right) && 
-				(y >= Top) && (y < Bottom));
+			return ((x >= Left) && (x <= Right) && 
+				(y >= Top) && (y <= Bottom));
 		}
 		
 		public bool IntersectsWith (Rectangle r)


lytico:
this needs more discussion.

see:
https://github.com/mono/mono/blob/master/mcs/class/System.Drawing/System.Drawing/Rectangle.cs
<pre>
public bool Contains (int x, int y)
{
    return ((x >= Left) && (x < Right) && 
	(y >= Top) && (y < Bottom));
 }
</pre>


microsofts system.drawing implementation:
<pre>
    public bool Contains(Rectangle rect)
    {
      if (this.X <= rect.X && rect.X + rect.Width <= this.X + this.Width && this.Y <= rect.Y)
        return rect.Y + rect.Height <= this.Y + this.Height;
      return false;
    }
</pre>

see also:
https://github.com/mono/xwt/pull/26
where i proposed the implemention following the mono system drawing implementation
in favour to @slluis implementions

@hwthomas any ideas about that?

lytico:
https://github.com/mono/mono/blob/master/mcs/class/WindowsBase/System.Windows/Rect.cs

<pre>

		public double Right { 
			get { return _x + _width; }
		}

		public double Bottom { 
			get { return _y + _height; }
		}

		public bool Contains (Rect rect)
		{
			if (rect.Left < this.Left ||
			    rect.Right > this.Right)
				return false;

			if (rect.Top < this.Top ||
			    rect.Bottom > this.Bottom)
				return false;

			return true;
		}

		public bool Contains (double x, double y)
		{
			if (x < Left || x > Right)
				return false;
			if (y < Top || y > Bottom)
				return false;

			return true;
}
</pre>

lytico:
closed to compare more implementation

hwthomas:
The Xwt Drawing model is closely based on the Cairo one, and as such is vector rather than pixel-based.  When applied to a pixel-based display, the X and Y coordinates may be thought of as defining a rectangular grid, with the pixels lying between the lines of this grid (as described in my Drawing notes in the Xwt Wiki).  Although Cairo doesn't have the 'Contains' function, I would expect a Rectangle (X, Y, Width, Height) to contain itself (ie its Top, Left, Bottom, and Right Points) and all interior Points (x,y) which satisfy X <= x <= (X+Width) and Y <= y <= (Y+Height).  Consequently, I'd agree with your proposed change to the Xwt Rectangle.Contains (Point p) code, though whether this throws up any other inconsistencies I'm not sure!   Incidentally, System.Drawing is pixel-based and not the same as Cairo, so its implementation isn't necessarily useful for comparison.
Hope this helps.

PR #622, opened 2016-09-22
[Gtk][DatePicker] Avoid use of Linq where possible

Therzok:


Therzok:
Nevermind, the generated code is O(1).

PR #627, opened 2016-11-15
L10n

elsupergomez:
Use Mono.Posix for Catalog to translate string.
This add Mono.Unix using and encapsulate string with Catalog.GetString()
In some cases use string.Format() to replace string concatenation.

### DIFF
--- a/Xwt/Xwt.Backends/BackendHost.cs
+++ b/Xwt/Xwt.Backends/BackendHost.cs
@@ -27,6 +27,7 @@ using System;
 using System.Collections.Generic;
 using System.Reflection;
 using System.ComponentModel;
+using Mono.Unix;
 
 
 namespace Xwt.Backends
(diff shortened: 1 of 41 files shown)


Therzok:
cc @slluis @sevoku Wouldn't it be better to use ResX for localization on new projects?

Therzok:
That way we do not depend on Mono.Unix on Windows too :)

sevoku:
ResX would be better, but I think that its even better to keep the Xwt translation API as simple and generic as possible, allowing to use any translation technique. See my proposal in https://github.com/mono/xwt/pull/629.

elsupergomez:
I like to know how use Xwt translation API for translate, as a example, to Spanish.

sevoku:
@elsupergomez: I've merged #629. The API is really simple, you need to implement the `ITranslationCatalog` (https://github.com/mono/xwt/blob/master/Xwt/Xwt/ITranslationCatalog.cs#L29-L33) interface and assign your implementation to the new `Application.TranslationCatalog` property (https://github.com/master/xwt/blob/master/Xwt/Xwt/Application.cs#L66).

elsupergomez:
It is really simple. I like it.

PR #629, opened 2016-12-05
Add translation support

sevoku:
This PR makes internal Xwt strings translatable though a custom `Xwt.ITranslationCatalog` implementation in `Application.TranslationCatalog`. The default implementation simply returns the same unteranslated string.

(closes #627)

### DIFF
--- a/Xwt.Gtk.Mac/GtkMacOpenFileDialogBackend.cs
+++ b/Xwt.Gtk.Mac/GtkMacOpenFileDialogBackend.cs
@@ -46,7 +46,7 @@ namespace Xwt.Gtk.Mac
 			if (!string.IsNullOrEmpty (initialFileName))
 				this.DirectoryUrl = new NSUrl (initialFileName, true);
 
-			this.Prompt = "Select File" + (multiselect ? "s" : "");
+			this.Prompt = Application.TranslationCatalog.GetPluralString("Select File", "Select Files", multiselect ? 2 : 1);
 		}
 
 		public bool Run (IWindowFrameBackend parent)
--- a/Xwt.Gtk/Xwt.GtkBackend/GtkAlertDialog.cs
+++ b/Xwt.Gtk/Xwt.GtkBackend/GtkAlertDialog.cs
@@ -123,7 +123,7 @@ namespace Xwt.GtkBackend
 			}
 			
 			if (message.AllowApplyToAll) {
-				Gtk.CheckButton check = new Gtk.CheckButton ("Apply to all");
+				Gtk.CheckButton check = new Gtk.CheckButton (Application.TranslationCatalog.GetString("Apply to all"));
 				this.AddContent (check, false, false, 0);
 				check.Toggled += delegate {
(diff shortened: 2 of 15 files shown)


Xwt.Gtk.Mac/GtkMacOpenFileDialogBackend.cs
```diff
@@ -46,7 +46,7 @@ public void Initialize (System.Collections.Generic.IEnumerable<FileDialogFilter>
 			if (!string.IsNullOrEmpty (initialFileName))
 				this.DirectoryUrl = new NSUrl (initialFileName, true);
 
-			this.Prompt = "Select File" + (multiselect ? "s" : "");
+			this.Prompt = Application.TranslationCatalog.GetPluralString("Select File", "Select Files", multiselect ? 2 : 1);
```
Therzok: Don't we have the real selection count?

sevoku: we could use a NSOpenSavePanelDelegate to track selection changes (same in https://github.com/sevoku/xwt/blob/add-translation-support/Xwt.XamMac/Xwt.Mac/FileDialogBackend.cs#L58). And we could also add some logic to Gtk/Wpf.

But actually I don't really like the plurals here, cause alsmost no other App does this. I think something more generic like "Select" (without "File"/"Folder") would fit better. Its even worse in Gtk (https://github.com/sevoku/xwt/blob/add-translation-support/Xwt.Gtk/Xwt.GtkBackend/FileDialogBackend.cs#L65-L81) where we use generic "non"-plural stock ids for other commands and folder selection/creation is always "Select Folder".

sevoku: Lets merge this as it is, if we like the approach and think about translation optimization and plurals separately.

Xwt.Gtk/Xwt.GtkBackend/GtkAlertDialog.cs
```diff
@@ -123,7 +123,7 @@ public GtkAlertDialog (ApplicationContext actx, MessageDescription message)
 			}
 			
 			if (message.AllowApplyToAll) {
-				Gtk.CheckButton check = new Gtk.CheckButton ("Apply to all");
+				Gtk.CheckButton check = new Gtk.CheckButton (Application.TranslationCatalog.GetString("Apply to all"));
```
elsupergomez: I have an error with conflict Xwt.Application with Gtk.Application. The line 126 should be:

Gtk.CheckButton check = new Gtk.CheckButton (Xwt.Application.TranslationCatalog.GetString("Apply to all"));

Adding an explicit reference to a Xwt.Application

sevoku: Oh thanks, and sorry! Fixed: https://github.com/mono/xwt/commit/f5f9a7449d68f4362aecf0ed800d336ca50fe0fd

PR #630, opened 2016-12-30
Add HasFocus property to Window.

konrad-kruczynski:
Implemented on all backends. Also some code of Mac backend that seemed to be a
leftover was removed.


sevoku:
LGTM, Thanks!

PR #640, opened 2017-02-09
Apply Xwt Translation API to exception texts

elsupergomez:
Signed-off-by: Claudio Rodrigo Pereyra Diaz <claudiorodrigo@pereyradiaz.com.ar>


slluis:
Exception texts are for developers, not for end users. They don't need to be translated.

PR #649, opened 2017-03-03
Add WindowFrame.BackgroundColor property

sevoku:


slluis:
I feel this belongs to Window, not WindowFrame.

sevoku:
Hhm, but its the same as Opacity. And depending on the toolkit the BG color is applied on the titlebar and window border as well (Cocoa does it definitely).

slluis:
I guess WindowFrame is not a good name. I added WindowFrame because I needed to represent a window which may have not been created by XWT. For example, when embedding an XWT widget inside an existing Gtk or Cocoa UI, the widget will still have a parent window, but it can't be an XWT.Window instance because that window is not under control of XWT, so it is a WindowFrame. So WindowFrame is intended to be a very simple representation of a Window, with little functionality. I guess in some cases it is not 100% clear whether a functionality belongs to WindowFrame or Window. For layouting properties such as Padding it is obvious it belongs to Window, since it requires the cooperation of XWT content widgets.

If GTK, Cocoa and WPF support BackgroundColor directly, I guess it could go into WindowFrame too, but I have the feeling that having it in Window makes more sense. It doesn't seem correct to allow modifying the content of a Window which you don't control. Opacity is kind of an edge case, since it is usually implemented by the window manager, and it has effect on the windows as a whole. It does change the window content, it just changes the way it is presented to the user (like changing visibility).

sevoku:
Ok, that makes sense. Lets move it to Window then.

sevoku:
@slluis done

PR #652, opened 2017-03-04
Add PopupWindow

sevoku:
This PR adds an `Xwt.PopupWindow` window (replacing the dummy Xwt.PopupWindow).
The PopupWindow is a special window type (based on `Xwt.Window`) which can be used to show custom Panels, custom Tooltips, custom Menus or highly customized windows.

By default a PopupWindow is a so called Utility Window, but it can also be created with an additional `PopupType` parameter to change the default behavior of the Popup:
* `PopupType.Utility`:
   Default Utility window (Cocoa: `NSPanel`, Gtk/Wpf: default window with utility style)
* `PopupType.Tooltip`:
   A special undecorated utility/popup window with support for background colors with alpha channel.
   Depending on the toolkit capabilities it may have a different pointer handling or animation.
* `PopupType.Menu`:
   In most cases same as `PopupType.Tooltip` but tells the WM to treat the window like a menu (depends on the WM to handle the hint).

**TODO:** `CanGetFocus` and `HasFocus` support, important for interactive tooltips.

### DIFF
--- a/TestApps/Samples/MainWindow.cs
+++ b/TestApps/Samples/MainWindow.cs
@@ -135,6 +135,7 @@ namespace Samples
 
 			var windows = AddSample (null, "Windows", typeof(Windows));
 			AddSample (windows, "Message Dialogs", typeof(MessageDialogs));
+			AddSample (windows, "Popup Windows", typeof(PopupWindows));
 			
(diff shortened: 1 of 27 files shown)


slluis:
What about having a Window subclass for each of those windows types? "PupupWindow" is not really very meaningful.

PR #658, opened 2017-03-13
[Mac] fix Label alignment with Label.Markup

sevoku:
NSTextField.Alignment has no effect if
NSTextField.AttributedString is being used
for Markup. Alignment needs to be added to
the NSAttributedString with an appropriate
NSParagraphStyle.

### DIFF
--- a/Xwt.XamMac/Xwt.Mac/LabelBackend.cs
+++ b/Xwt.XamMac/Xwt.Mac/LabelBackend.cs
@@ -27,6 +27,7 @@
 using System;
 using AppKit;
 using CoreGraphics;
+using Foundation;
 using ObjCRuntime;
 using Xwt.Backends;
 
@@ -96,6 +97,10 @@ namespace Xwt.Mac
 				Wrap = WrapMode.Character;
 			Widget.AllowsEditingTextAttributes = true;
 			Widget.AttributedStringValue = text.ToAttributedString ();
+
+			if (TextAlignment != Alignment.Start)
+				SetAlignmentAttribute ();
+
 			ResetFittingSize ();
 		}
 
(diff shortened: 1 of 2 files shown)


bratsche:
👍

Xwt.XamMac/Xwt.Mac/LabelBackend.cs
```diff
@@ -96,6 +97,10 @@ public void SetFormattedText (FormattedText text)
 				Wrap = WrapMode.Character;
 			Widget.AllowsEditingTextAttributes = true;
 			Widget.AttributedStringValue = text.ToAttributedString ();
```
Therzok: Maybe we can avoid double drawing of the string if we add an extension in MacUtil.cs:

```
public static NSMutableAttributedString ToMutableAttributedString (this FormattedText ft)
{
    return (NSMutableAttributedString)ToAttributedString (ft);
}

And use this before calling SetAlignmentAttribute.

Then, `SetAlignmentAttribute` would become
```
void SetAlignmentAttribute (NSMutableAttributedString ns)
{
 			if (ns == null)
 				return;
 			
 			ns.BeginEditing ();
 			var r = new NSRange (0, ns.Length);
 			ns.RemoveAttribute (NSStringAttributeKey.ParagraphStyle, r);
 			var pstyle = NSParagraphStyle.DefaultParagraphStyle.MutableCopy () as NSMutableParagraphStyle;
 			pstyle.Alignment = Widget.Alignment;
 			ns.AddAttribute (NSStringAttributeKey.ParagraphStyle, pstyle, r);
 			ns.EndEditing ();
 			Widget.AttributedStringValue = ns;
 		}
```

Therzok: By double drawing, I mean not set Widget.AttributedStringValue twice.

sevoku: I think we move the whole thing to Util

Therzok:
Much better.

PR #660, opened 2017-03-16
Accessibility Support

sevoku:
This PR adds a basic Accessibility API which can be accessed through `Widget.Accessible`.

Detailed Description follows ...


Xwt/Xwt.Accessibility/IAccessibleBackend.cs
```diff
@@ -0,0 +1,63 @@
...
+{
+	public interface IAccessibleBackend : IBackend
+	{
+		void InitializeAccessible (IAccessibleEventSink eventSink);
+
+		bool IsAccessible { get; set; }
+
+		string AccessibleLabel { get; set; }
+
+		string AccessibleTitle { get; set; }
```
slluis: This is IAccessibleBackend, there is no need to add "Accessible" in every member of the interface.

Xwt/Xwt.Accessibility/Accessible.cs
```diff
@@ -0,0 +1,171 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+using System;
+using Xwt.Backends;
+
+namespace Xwt.Accessibility
+{
+	[BackendType (typeof (IAccessibleBackend))]
+	public sealed class Accessible : IAccessibleEventSink
```
slluis: I don't want frontend classes to implement backend interfaces. Events should come through the widget backend host.

Xwt/Xwt.Accessibility/IAccessibleBackend.cs
sevoku: done :)

Xwt/Xwt.Accessibility/Accessible.cs
sevoku: done :)

slluis: By widget backend host I mean the host that we already have in the Widget class. We already discussed about this. The widget host can implement IAccessibleEventSink and internally route the events to Accessible. In this way we can avoid having a yet another host instance for each widget.

sevoku: Yes, but this approach allows us to wrap external accessibility objects out of the box, which is impossible if we route everything through the `Widget` class.

sevoku: Also we'll have an other host instance only for each `Widget.Accessible` instance which is created on demand. If the user doesn't use the API, it won't be created.

slluis: You can have a different backend host when wrapping external objects. Impossibility solved! Anyway, keep it as is for now, not a big issue.

Xwt.Gtk/Xwt.GtkBackend/AccessibleBackend.cs
```diff
@@ -0,0 +1,154 @@
...
+				return widget.Accessible.Description;
+			}
+			set {
+				widget.Accessible.Description = value;
+			}
+		}
+
+		public string Label {
+			get {
+				return widget.Accessible.Name;
```
iainx: The Name property in Atk isn't a label, but an identifier such as "MyExcellentDialog.AdditionButton"

It is called accessibilityIdentifier in Cocoa.

Xwt/Xwt.Accessibility/IAccessibleBackend.cs
```diff
@@ -0,0 +1,65 @@
...
+
+		bool IsAccessible { get; set; }
+
+		string Label { get; set; }
+
+		string Title { get; set; }
+
+		string Description { get; set; }
+
+		string Value { get; set; }
```
iainx: In Cocoa a value can be any object. For a text entry it is a string, for a progress bar it's a number.

Xwt.Gtk/Xwt.GtkBackend/AccessibleBackend.cs
sevoku: Ok, then it should be just a dummy maybe?

Xwt/Xwt.Accessibility/IAccessibleBackend.cs
sevoku: Can it also be a complex object? As long as its convertible to/from a string (like a number), we can keep the string abstraction for simplicity and let each backend do its specific conversion.

Xwt.Gtk/Xwt.GtkBackend/AccessibleBackend.cs
iainx: yeah, could work

Xwt/Xwt.Accessibility/IAccessibleBackend.cs
iainx: I'm not really sure, the documentation doesn't say.
I've just checked what NSColorWell returns, but it just returns a string. Lets say it can either return an NSString or NSNumber until we find out otherwise?

PR #661, opened 2017-03-20
[Xwt] Add a faster way of getting the image size

alanmcgovern:
We want the image size without loading and parsing the entire image


alanmcgovern:
👍

PR #662, opened 2017-03-28
Add ListViewColumn and CellView expansion

sevoku:
This patch adds expansion support for ListView/TreeView columns and cells.
New API:
* `ListViewColumn.Expands`: the column expands to fill all available space inside the view.
* `CellView.Expands`: the cell view expands horizontally to fill all available space inside its column. Due to toolkit limitations CellView expanding can be only set upon creation, `CellView.Expands` is read only.

TODO: Wpf column expansion

### DIFF
--- a/TestApps/Samples/Samples/ListView1.cs
+++ b/TestApps/Samples/Samples/ListView1.cs
@@ -21,8 +21,9 @@ namespace Samples
 			ListStore store = new ListStore (name, icon, text, icon2, progress);
 			list.DataSource = store;
 			list.Columns.Add ("Name", icon, name);
-			list.Columns.Add ("Text", icon2, text);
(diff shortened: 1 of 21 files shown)


slluis:
I feel it would be more consistent if the expand argument is specified when adding the cell to the cellview collection. `CellViewCollection` could have overloads for Add and Insert that take an `expands` argument (which could still be internally stored in CelllView).

sevoku:
Yeah, that's a good idea! Done: f745dc8
The CellView constructor parameter is still useful I think, or should we remove it now?

slluis:
I would remove it now.

sevoku:
ok, done

garuma:
Merged since we need this for one of our feature work.

PR #668, opened 2017-04-14
[Mac] Fixed using Popover with `NSView`.

viniciusjarina:
* Added a overload to `IPopoverBackend.Show` receiving an object.
We can’t wrap and unwrap the NSView to be used with the Popover, because this will change the parent of the window when we unwrap it, invalidating the `NSView.Window`, making NSPopover throw a NSException, due the missing NSWindow inside the NSView.
* This patch pass the NSView all the way down, without wrap it inside a EmbeddedNativeWidget, this way avoiding the NSView to lost the NSWindow.

### DIFF
--- a/Xwt.Gtk/Xwt.GtkBackend/PopoverBackend.cs
+++ b/Xwt.Gtk/Xwt.GtkBackend/PopoverBackend.cs
@@ -245,9 +245,23 @@ namespace Xwt.GtkBackend
 		{
 		}
 
+		public void Show (Xwt.Popover.Position orientation, object reference, Xwt.Rectangle positionRect, Widget child)
+		{
+			var nativeReference = reference as Gtk.Widget;
+			if (nativeReference == null)
+				return;
+			Show (orientation, nativeReference, positionRect, child);
+		}
+
 		public void Show (Xwt.Popover.Position orientation, Xwt.Widget reference, Xwt.Rectangle positionRect, Widget child)
 		{
-			popover.Content = (Gtk.Widget)((WidgetBackend)Toolkit.GetBackend (child)).NativeWidget;
+			var nativeReference = (Gtk.Widget)((WidgetBackend)Toolkit.GetBackend (child)).NativeWidget;
+			Show (orientation, nativeReference, positionRect, child);
+		}
+
+		public void Show (Xwt.Popover.Position orientation, Gtk.Widget reference, Xwt.Rectangle positionRect, Widget child)
+		{
+			popover.Content = reference;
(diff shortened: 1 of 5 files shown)


Xwt.Gtk/Xwt.GtkBackend/PopoverBackend.cs
```diff
@@ -262,10 +276,10 @@ public void Show (Xwt.Popover.Position orientation, Xwt.Widget reference, Xwt.Re
 
 			popover.Hidden += (o, args) => sink.OnClosed ();
 
-			var screenBounds = reference.ScreenBounds;
+			var screenBounds = reference.Allocation;
```
sevoku: Why is this change needed? `Widget.Allocation` is not reliable, especially for custom drawn widgets and widgets without own `Gdk.Window` and/or Gtk widgets inside an environment mixed with Cocoa. This is the reason why I used `ScreenBounds` here. Additionally the coords are used to `Move` the Popover to the destination screen coordinates, while `Allocation` is relative to the parent or window in most cases.

Xwt.XamMac/Xwt.Mac/PopoverBackend.cs
```diff
@@ -200,6 +200,22 @@ public void DisableEvent (object eventId)
...
 		}
 
+		public void Show (Popover.Position orientation, object reference, Rectangle positionRect, Widget child)
+		{
+			var refView = reference as NSView;
+			if (refView == null)
+				return;
+
+			// If the rect is empty, the coordinates of the rect will be ignored.
+			// Set the width and height, for the positioning to function correctly.
```
sevoku: If the rectangle is empty, Xwt should use the target widget bounds. At least this is how its supposed to be and its the default Cocoa behavior. If you need to specify coordinates relative to the target widget, you should use a non-empty rectangle around the coords and the popup should align to the center of the specified edge of this rectangle.

sevoku:
> We can’t wrap and unwrap the NSView to be used with the Popover, because this will change the parent of the window when we unwrap it, invalidating the NSView.Window, making NSPopover throw a NSException, due the missing NSWindow inside the NSView.

Can't we modify the wrapping and fix the NSWindow thing there? With an additional param to `WrapWidget` maybe?

viniciusjarina:
Hi @sevoku I think this is a great idea and would simplify things.

I did the change I did just a proof of concept to be sure this is what was causing the native crash on Mac. 

But I think would be easier an no-touch Wrap/Unwrap API 👍

PR #671, opened 2017-04-15
[Xwt,Gtk,WPF] Add support for NewText handling in TextCellView

Therzok:
I have not tested the WPF backend, but pretty sure the Gtk backend is right.

This adds support for checking whether or not to modify a text cell's text based on the text being set.

### DIFF
--- a/Xwt.WPF/Xwt.WPFBackend/TextEntryBackend.cs
+++ b/Xwt.WPF/Xwt.WPFBackend/TextEntryBackend.cs
@@ -225,7 +225,7 @@ namespace Xwt.WPFBackend
 				Context.InvokeUserCode (EventSink.OnActivated);
 		}
 
-		private void OnTextChanged (object s, TextChangedEventArgs e)
+		private void OnTextChanged (object s, System.Windows.Controls.TextChangedEventArgs e)
 		{
 			Context.InvokeUserCode (EventSink.OnChanged);
 		}
--- a/Xwt/Xwt/TextChangedEventArgs.cs
+++ b/Xwt/Xwt/TextChangedEventArgs.cs
@@ -2,9 +2,9 @@
 // TextChangedEventArgs.cs
 //
 // Author:
-//       therzok <>
+//       Marius Ungureanu <maungu@microsoft.com>
 //
-// Copyright (c) 2017 ${CopyrightHolder}
+// Copyright (c) 2017 Microsoft Corporation
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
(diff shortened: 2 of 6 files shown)


Xwt.WPF/Xwt.WPFBackend/TextEntryBackend.cs
```diff
@@ -225,7 +225,7 @@ private void OnActivated(object sender, System.Windows.Input.KeyEventArgs e)
 				Context.InvokeUserCode (EventSink.OnActivated);
 		}
 
-		private void OnTextChanged (object s, TextChangedEventArgs e)
+		private void OnTextChanged (object s, System.Windows.Controls.TextChangedEventArgs e)
```
sevoku: is this needed?

Xwt/Xwt/TextChangedEventArgs.cs
```diff
@@ -0,0 +1,37 @@
+﻿//
+// TextChangedEventArgs.cs
+//
+// Author:
+//       therzok <>
```
sevoku: add your name and copyright to the header

sevoku:
WPF is tricky, it uses data bindings and the `TextChanged` event (https://github.com/mono/xwt/blob/ed5e1a30c535ddda82e661cd1033dff6d6d7769e/Xwt.WPF/Xwt.WPFBackend.CellViews/TextCellViewBackend.cs#L36) is raised after the data has been changed already. I tried to use a custom TextBox and override OnTextChanged, but it has the same problem (approach I used for CheckBox in https://github.com/mono/xwt/commit/8f23789090ea2bb54f1c962e0fcbdde512ccdbc1#diff-ddcae0ee4362cba10902c4ecfaf62894R66 doesn't work here). We handle `PreviewTextInput`, but it would need much more logic around it to handle backspace etc.

And does it build with XamMac?

Xwt.WPF/Xwt.WPFBackend/TextEntryBackend.cs
Therzok: Not sure.

Xwt/Xwt/TextChangedEventArgs.cs
Therzok: Will fix.

Therzok:
XamMac doesn't have this implemented at all at the moment, so yes, it builds.

Maybe we can use PreviewTextInput for this, but that's raised on any keypress, not just text changed. Maybe TextChangedEventArgs could have both `oldText` and `newText`, and we keep an internal state in PreviewTextInput and put both of them in the event args?

Therzok:
About WPF, can't we cache each TextChanged call argument and provide the old value? I don't suppose the value can change without notification.

sevoku:
@Therzok I've added a fix for WPF: 09bb47e
It simply uses the Text Binding to reset the text to its original value.

Therzok:
Can we merge this and publish a new nuget? Can't wait to finish the resx editor.

PR #675, opened 2017-04-18
[Mac] Fix RichTextView wrapping inside ScrollView

sevoku:
If the enclosing ScrollViews horizontal scrolling
is disabled, GetPreferredSize should not use the initial
text width in order to allow the text view to wrap
the text.


sevoku:
This patch can be verified by setting `HorizontalScrollPolicy = ScrollPolicy.Never` in https://github.com/mono/xwt/blob/master/TestApps/Samples/Samples/Markdown.cs#L153.
Without the patch the `MarkdownView` will request a width to fit the text without wrapping, while it should wrap to fit the available width.

PR #676, opened 2017-04-18
[Mac] Fixed crash using Popover with `NSView`.

viniciusjarina:
* Added a parameter to `WrapWidget` to do not reparent the nativeView.
We can’t wrap and unwrap the NSView to be used with the Popover, because this will change the parent of the window when we unwrap it, invalidating the `NSView.Window`, making NSPopover throw a NSException, due the missing NSWindow inside the NSView.
* This add a option to do not reparent the native view when Wrap/Unwrap, and get the (untouched) native view back on `Show`.

Previous related PR: https://github.com/mono/xwt/pull/668

### DIFF
--- a/Xwt.Gtk/Xwt.GtkBackend/EmbeddedWidgetBackend.cs
+++ b/Xwt.Gtk/Xwt.GtkBackend/EmbeddedWidgetBackend.cs
@@ -34,7 +34,7 @@ namespace Xwt.GtkBackend
 		{
 		}
 
-		public void SetContent (object nativeWidget)
+		public void SetContent (object nativeWidget, bool reparent)
 		{
 			if (nativeWidget is Gtk.Widget) {
 				Widget = (Gtk.Widget)nativeWidget;
--- a/Xwt.WPF/Xwt.WPFBackend/EmbedNativeWidgetBackend.cs
+++ b/Xwt.WPF/Xwt.WPFBackend/EmbedNativeWidgetBackend.cs
@@ -12,7 +12,7 @@ namespace Xwt.WPFBackend
 {
 	class EmbedNativeWidgetBackend: WidgetBackend, IEmbeddedWidgetBackend
 	{
-		public void SetContent(object nativeWidget)
+		public void SetContent(object nativeWidget, bool reparent)
 		{
 			if (nativeWidget is FrameworkElement)
 				Widget = (FrameworkElement)nativeWidget;
--- a/Xwt.XamMac/Xwt.Mac/EmbedNativeWidgetBackend.cs
+++ b/Xwt.XamMac/Xwt.Mac/EmbedNativeWidgetBackend.cs
@@ -1,4 +1,4 @@
-using AppKit;
+﻿using AppKit;
 using Xwt.Backends;
 
 namespace Xwt.Mac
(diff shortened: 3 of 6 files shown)


chkn:
What is a case when you would want to pass `reparent: true`? Would it make more sense to just remove the reparenting behavior entirely?

viniciusjarina:
@chkn I think not, all other parts of Xwt rely on this behavior, this is a corner case scenario where reparenting was causing the crash with `NSPopover` 😉

Xwt.XamMac/Xwt.Mac/PopoverBackend.cs
```diff
@@ -203,7 +203,11 @@ public void DisableEvent (object eventId)
 		public void Show (Popover.Position orientation, Widget referenceWidget, Rectangle positionRect, Widget child)
 		{
 			var refBackend = Toolkit.GetBackend (referenceWidget) as IWidgetBackend;
-			NSView refView = (refBackend as ViewBackend)?.Widget;
+
+			NSView refView = (refBackend as EmbedNativeWidgetBackend)?.InnerView;
```
sevoku: Why is this required? Also it could break event handling, for events being attached after the widget has been added to the Xwt tree. I've got the feeling that we handle a product specific corner case here... and should make sure that we don't break other potential use cases.

Xwt.XamMac/Xwt.Mac/EmbedNativeWidgetBackend.cs
```diff
@@ -1,30 +1,37 @@
...
 	{
 		NSView innerView;
+		bool reparent;
 
 		public EmbedNativeWidgetBackend ()
 		{
 
 		}
 
+		public NSView InnerView {
```
sevoku: Why is it necessary to acces the inner widget on XamMac backend level? Can't `Widget.Subview` be used to access the inner widget?

Xwt.Gtk/Xwt.GtkBackend/EmbeddedWidgetBackend.cs
```diff
@@ -34,7 +34,7 @@ public EmbeddedWidgetBackend ()
 		{
 		}
 
-		public void SetContent (object nativeWidget)
+		public void SetContent (object nativeWidget, bool reparent)
```
sevoku: The Gtk backend reparents the widget automatically by default: https://github.com/mono/xwt/blob/master/Xwt.Gtk/Xwt.GtkBackend/WidgetBackend.cs#L105
Since we introduce a new API here, I think we should throw an `NotImplementedException` in case `reparent: false`, or add the Gtk specific implementation for the new API.

Xwt.WPF/Xwt.WPFBackend/EmbedNativeWidgetBackend.cs
```diff
@@ -12,7 +12,7 @@ namespace Xwt.WPFBackend
 {
 	class EmbedNativeWidgetBackend: WidgetBackend, IEmbeddedWidgetBackend
 	{
-		public void SetContent(object nativeWidget)
```
viniciusjarina: 👍

PR #684, opened 2017-04-28
Always use native backend for Clipboard

chkn:
Without this, `Clipboard.ContainsData` can return different results depending on which backend it is invoked with.

The only caveat I see with this approach is that any `Image` object returned from the Clipboard API will now have the native backend, which might be different than where it will be consumed. But since we have multi-toolkit support for images, I don't think this should be an issue.

### DIFF
--- a/Xwt/Xwt/Clipboard.cs
+++ b/Xwt/Xwt/Clipboard.cs
@@ -33,10 +33,18 @@ namespace Xwt
 {
 	public static class Clipboard
 	{
+		// Cache the toolkit the first time we access the Clipboard and then always use
+		//  the same one, otherwise we will get inconsistent results from APIs like ContainsData<T>().
+		static Toolkit toolkit;
+
 		static ClipboardBackend Backend {
-			get { return Toolkit.CurrentEngine.ClipboardBackend; }
+			get {
+				if (toolkit == null)
+					toolkit = Toolkit.CurrentEngine;
+				return toolkit.ClipboardBackend;
+			}
 		}
-		
+
 		public static void Clear ()
 		{
 			Backend.Clear ();


sevoku:
I'm not sure that its enough just to change the toolkit here. It makes it impossible for the consumer to take control of the toolkit and it will break pure Gtk driven applications (not initializing native toolkits and shipping `Xwt.Gtk.dll` only.) on platforms other than Linux.

One possible workaround could be to set `Toolkit.NativeEngine` to `Toolkit.CurrentEngine` in case of missing native engine. But this would change the meaning of it. TBH it should always return the native toolkit or null (atm it throws, hence needs to be fixed).

An other option would be to add a `Clipboard.Toolkit { get; set; }` property, which would be `Toolkit.CurrentEngine` by default, but can be set to any other toolkit.

chkn:
> ...it will break pure Gtk driven applications (not initializing native toolkits and shipping `Xwt.Gtk.dll` only.) on platforms other than Linux. One possible workaround could be to set `Toolkit.NativeEngine` to `Toolkit.CurrentEngine` in case of missing native engine.

That is the current behavior, so this is not an issue (https://github.com/mono/xwt/blob/master/Xwt/Xwt/Toolkit.cs#L104-L105)

> An other option would be to add a `Clipboard.Toolkit { get; set; }` property, which would be `Toolkit.CurrentEngine` by default, but can be set to any other toolkit.

This does not address the original issue, which is that `Clipboard.ContainsData<MyType>()` is returning different values depending on the backend. To fix this, I only see 2 options:

1) Always use the same backend for the Clipboard (this PR)
2) Fix all the backends so that they store/query for custom types in a compatible way

Option (1) seems like a more maintainable route. If we do not want to use `Toolkit.NativeEngine`, we could also capture the initial value of `Toolkit.CurrentEngine` when the Clipboard is first accessed and then use that for subsequent accesses.

sevoku:
>> An other option would be to add a Clipboard.Toolkit { get; set; } property, which would be
>> Toolkit.CurrentEngine by default, but can be set to any other toolkit.

> This does not address the original issue, which is that Clipboard.ContainsData<MyType>() is
> returning different values depending on the backend.

Sorry, could you explain why it won't address the original issue?

```c#
public static Toolkit Toolkit { get; set; }

static ClipboardBackend Backend {
	get { return (Toolkit ?? Toolkit.CurrentEngine).ClipboardBackend; }
}
```
this ^^ is what I propose. By default it would always use the current engine. To fix the issue, you just need to set `Clipboard.Toolkit = Toolkit.NativeEngine` once at startup.

chkn:
> By default it would always use the current engine. To fix the issue, you just need to set Clipboard.Toolkit = Toolkit.NativeEngine once at startup.

This is what is broken. It is broken by default for any multi-toolkit app. What you propose still requires the user to know that the default behavior is broken and do something about it. Our copy/paste handling was probably broken for ages and we didn't know, so I wouldn't expect new Xwt users to "just know" that they need to set this setting.

chkn:
@sevoku How about this?

```csharp
public static Toolkit Toolkit { get; set; }

static ClipboardBackend Backend {
	get {
		if (Toolkit == null)
			Toolkit = Toolkit.CurrentEngine;
		return Toolkit.ClipboardBackend;
	}
}
```

sevoku:
The default behaviour everywhere is that `Toolkit.CurrentEngine` is the default and currently used toolkit. This is how it was intended and we make use of it everywhere (I used it recently to implement cross-toolkit/native popups for MD where `Desktop` has exactly the same problem). By default all static classes have no "fixed" toolkit, and we should keep that pattern for consistancy.

In case of the `Clipboard` class it doesn't make a difference whether you use Gtk or XamMac, its only important that you always use the same engine, but this is only true for our current Gtk/XamMac backend implementations, and may change with new toolkits (for example) in the future. What you propose is to limit the API, or to change it in a way that at least its usage differs compared to i.e. `Desktop`. So without looking into its implmentation, lets see it like a black box, it doesn't allow to switch backends in a controlled environment by invoking it with `Engine.Invoke`, which is possible everywhere else.

chkn:
Then I guess we just have different points of view. To me, I have certain expectations about how the clipboard works-- that there is a single one and it is globally consistent-- and the backend is just an implementation detail that I don't care about as long as the clipboard works that way.

But if it is to be "by design" that `Clipboard.ContainsData<MyType>()` can return different results depending on the backend, then we can add your property, or I will grep through our code and add a toolkit invoke around all the clipboard stuff. But we've definitely had a bug in our code for *years* due to this and it was very hard to discover (the handlers were triggered with Gtk as the current toolkit when invoked from the system menu, while XamMac was the current toolkit when invoked from the context menu, so you couldn't copy using one menu and then paste using the other, but the same menu would work). I'm sure other people will have equally hard-to-detect bugs in their code due to this as well, so it is a bummer to leave them to fall into this trap.

sevoku:
> To me, I have certain expectations about how the clipboard works-- that there is a single one and it is globally consistent-- and the backend is just an implementation detail that I don't care about as long as the clipboard works that way.

In an ideal word, definitely true. I expected the same to be true for `Desktop`, but it doesn't work that way, because different stuff relies on it (like `Widget.ScreenBounds` for example), hence I need to invoke it with different toolkits. ATM nothing else inside Xwt itself relies on `Clipboard.Backend` to be the right one, but this may change in the future.

At the end its a API design decision. Consistancy across Xwt features vs. reliability by default. I think 99% of Xwt users don't really care about cross-toolkit environments and would never face that issue. And someone who really uses it in a way we do in MonoDevelop, should be aware of `Toolkit.CurrentEngine` limitations and knows about the requirement to invoke code in the right context. And it should be documented somewhere, but we have no cross-toolkit docs at all.

slluis:
For now, I would prefer a solution like this:

```csharp
static Toolkit toolkit;

static ClipboardBackend Backend {
	get {
		if (toolkit == null)
			toolkit = Toolkit.CurrentEngine;
		return toolkit.ClipboardBackend;
	}
}
```

This is a simple fix that solves the problem. It is true that with this change we lose the ability to invoke Clipboard methods for a specific backend, but right now this is not a problem. If this ever becomes a problem then we'll find the right solution.

`Engine.Invoke` was added because we have static methods that need to know which toolkit to use. However if Xwt can handle cross-tolkit issues transparently, it should do it. This was done for example to support cross-toolkit images without the need for `Engine.Invoke`. `Desktop` can be fixed to behave the same way no matter the toolkit backend.

Another approach for solving this problem could be: when calling Clipboard.SetData, call SetData in all loaded backends. But the above solution is more simple, and we can always implement this other solution if we need it.

chkn:
Thanks @slluis. I rewrote the commit using that approach.

PR #693, opened 2017-05-28
[Mac] Optimize image drawing context.

Therzok:
This commit works around code that reflects a lot. NSObject.Copy() does
reflection to find the intptr constructor of an object, then uses
ConstructorInfo.Invoke to construct the NSImage.

Since we don't operate with the actual NSImage in the case we don't
have a custom image, do not create a copy, as we only need to pass in
the CGImage we get. The AsCGImage does not do any reflection, so this
should turn into faster code.

### DIFF
--- a/Xwt.XamMac/Xwt.Mac/ContextBackendHandler.cs
+++ b/Xwt.XamMac/Xwt.Mac/ContextBackendHandler.cs
@@ -296,8 +296,6 @@ namespace Xwt.Mac
 
 			// Add the styles that have been globaly set to the context
 			img.Styles = img.Styles.AddRange (cb.Styles);
-			
-			NSImage image = img.ToNSImage ();
 
 			ctx.SaveState ();
 			ctx.SetAlpha ((float)img.Alpha);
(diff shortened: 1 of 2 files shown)


sevoku:
LGTM!

Xwt.XamMac/Xwt.Mac/ContextBackendHandler.cs
```diff
@@ -309,12 +307,14 @@ public override void DrawImage (object backend, ImageDescription img, Rectangle
 			ctx.TranslateCTM ((float)(destRect.X - (srcRect.X * rx)), (float)(destRect.Y - (srcRect.Y * ry)));
 			ctx.ScaleCTM ((float)rx, (float)ry);
 
+			NSImage image = img.ToNSImage (copyNSImage: false);
```
slluis: Since `copyNSImage: false` means returning the NSImage backend as is, it would be more simple to just do this operation here. That is, instead of creating a new overload of ToNSImage, just do `image = (NSImage)img.Backend` here.

Therzok: Can do.

Therzok:
Fixed!

PR #694, opened 2017-05-31
Perf opt

Therzok:


Xwt.XamMac/Xwt.Mac/ContextBackendHandler.cs
```diff
@@ -38,15 +38,15 @@ class CGContextBackend {
 		public CGContext Context;
 		public CGSize Size;
 		public CGAffineTransform? InverseViewTransform;
-		public Stack<ContextStatus> StatusStack = new Stack<ContextStatus> ();
-		public ContextStatus CurrentStatus = new ContextStatus ();
+		public ContextStatus CurrentStatus;
```
slluis: CurrentStatus should never be null. It holds the currently active pattern.

Therzok: Yes, sorry about this.

Xwt.XamMac/Xwt.Mac/ContextBackendHandler.cs
```diff
@@ -63,17 +63,19 @@ public override void Save (object backend)
...
+				Previous = ct.CurrentStatus,
+			};
 		}
 		
 		public override void Restore (object backend)
 		{
 			var ct = (CGContextBackend) backend;
 			ct.Context.RestoreState ();
-			ct.CurrentStatus = ct.StatusStack.Pop ();
+			if (ct.CurrentStatus != null) {
```
slluis: Now you should check `ct.CurrentStatus.Previous != null`

Therzok: *throws stuff at former self for not saving the file before git add -p*

PR #697, opened 2017-06-02
Bigleaks

Therzok:
@slluis this is critical for 15.3


sevoku:
Oh, I really don't know what I did here :( the init logic doesn't make any sense, perhaps I messed up some merge, or something. But this change definitely breaks the PopupWindow initialization.

PR #698, opened 2017-06-02
Fix PopupWindow initialization

sevoku:
This fixes the `PopupWindow`initialization by calling `IPopupWindowBackend.Initialize` before `IWindowFrameBackend.Initialize` and the backends should take care of the rest.


Therzok:
No more leaks.

PR #700, opened 2017-06-10
[A11y] Initial work getting GtkMac backend

iainx:


Xwt.Gtk.Mac/GtkMacAccessibleBackend.cs
```diff
@@ -0,0 +1,67 @@
...
+
+namespace Xwt.Gtk.Mac
+{
+	public class GtkMacAccessibleBackend : AccessibleBackend
+	{
+		public GtkMacAccessibleBackend ()
+		{
+		}
+
+		const string XamarinPrivateAtkCocoaNSAccessibilityKey = "xamarin-private-atkcocoa-nsaccessibility";
```
Therzok: Shouldn't we be registering this in MonoDevelop itself?

iainx: It's also in MonoDevelop, but Xwt doesn't depend on MD

Therzok: No, I mean registering this backend.

PR #702, opened 2017-06-12
[a11y] Set the role of the ImageBox to an Image

iainx:
On Gtk the ImageBox is a custom Gtk.DrawingArea so it needs to have an image role set for accessibility


iainx:
Fixes XBC 53555

PR #706, opened 2017-06-14
Create a sync context per toolkit

slluis:
Every toolkit now has its own synchronization context. This is necessary
to avoid unexpected toolkit switches when doing async operations.

Application.Invoke and TimeoutInvoke will now also dispatch the callbacks
in the context of the toolkit that made the call.


alanmcgovern:
👍

PR #712, opened 2017-07-20
Remove old warnings in Xwt and Xwt.Gtk

xecrets:
There are a few unnecessary warnings in Xwt and Xwt.Gtk which are just annoying. We're using Xwt from source currently and all of our other projects have zero warnings and will fail if any warnings are issued. It just seems like it's time to get rid of these few warnings.


slluis:
It looks ok, but can you please use tabs instead of spaces for indenting? thanks.

xecrets:
Sure. Tabs it is.

Therzok:
Thanks!

PR #715, opened 2017-08-25
[Gtk] Add support for prelit

decriptor:


slluis:
What is this needed for?

sevoku:
This indicates whether a cell is currently highlighted (mouse over) or not. But it works only for Gtk this way, hence we'll need a more sophisticated solution. I think its time for a cell state as a param to the Draw event. I'll need to think about it next week.

Xwt.Gtk/Xwt.GtkBackend.CellViews/CustomCellRenderer.cs
```diff
@@ -93,6 +95,14 @@ internal void EndDrawing ()
 				return base.HasFocus;
 			}
 		}
+
+		public override bool MouseOver {
+		 get {
```
sevoku: wrong formatting (spaces->tabs)

Xwt/Xwt/CellView.cs
```diff
@@ -206,6 +206,10 @@ void ICellViewFrontend.DetachBackend ()
 			get { return Backend.HasFocus; }
 		}
 
+		protected bool MouseOver {
```
sevoku: Wouldn't `IsHighlighted` be better?

PR #716, opened 2017-08-31
Add TextAlignment property to TextLayout

chkn:
This enables different text alignment when drawing multiline TextLayout (Width property set).

Includes Gtk, Mac, and WPF implementations.


slluis:
It would be nice to have some test code.

Xwt/Xwt.Drawing/TextLayout.cs
```diff
@@ -109,6 +110,10 @@ internal void InitForToolkit (Toolkit tk)
 					handler.SetWidth (Backend, width);
 				if (height != -1)
 					handler.SetHeight (Backend, height);
+				if (textAlignment != default(Alignment))
+					handler.SetAlignment (Backend, textAlignment);
+				if (textTrimming != default(TextTrimming))
```
sevoku: This doesn't seem to be related, or is it? Looks like it was just missing, then we should split it into an other commit (and bugfix PR).

chkn:
@sevoku @slluis I've added a couple unit tests and an example in the samples app.

sevoku:
Merged!

PR #717, opened 2017-09-04
Handle context menu

alanmcgovern:
I discussed with @sevoku a while back about exposing an API to allow us to cleanly support the macos specific behaviour of displaying context menu's when the user uses `control-click`.

Docs: https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/MenuList/Articles/HowMenusWork.html

Does this approach make sense to you guys? If so we'll finalise and merge it!

### DIFF
--- a/Xwt.XamMac/Xwt.Mac/Util.cs
+++ b/Xwt.XamMac/Xwt.Mac/Util.cs
@@ -240,6 +240,21 @@ namespace Xwt.Mac
 			}
 		}
 
+		public static bool TriggersContextMenu (this NSEvent theEvent)
+		{
+			if (theEvent.ButtonNumber == 1 &&
+					(NSEvent.CurrentPressedMouseButtons & 1 | NSEvent.CurrentPressedMouseButtons & 4) == 0) {
+				return true;
+			}
+
+			if (theEvent.ButtonNumber == 0 && (theEvent.ModifierFlags & NSEventModifierMask.ControlKeyMask) != 0 &&
+					(NSEvent.CurrentPressedMouseButtons & 2 | NSEvent.CurrentPressedMouseButtons & 4) == 0) {
+				return true;
+			}
+
+			return false;
+		}
+
 		public static NSImage ToNSImage (this ImageDescription idesc)
 		{
 			if (idesc.IsNull)
--- a/Xwt/Xwt/ButtonEventArgs.cs
+++ b/Xwt/Xwt/ButtonEventArgs.cs
@@ -80,6 +80,12 @@ namespace Xwt
 		/// </summary>
 		/// <value>The number of the presses of the same button.</value>
 		public int MultiplePress { get; set; }
+
+		/// <summary>
+		/// Gets or sets value indicating whether context menu should be shown
+		/// </summary>
+		/// <value><c>true</c> if this context menu should be shown; otherwise, <c>false</c>.</value>
+		public bool IsContextMenuTrigger { get; set; }
 	}
 }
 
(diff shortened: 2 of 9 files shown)


slluis:
This looks a bit too complicated. Couldn't the mac backend just set the `IsMenuContextTrigger` property like it does for other `ButtonEventArgs` properties?

alanmcgovern:
Do you want every place which instantiates a `ButtonEventArgs` object to explicitly set the correct value for this property?

slluis:
Yes, although it would be better to factorize ButtonEventArgs creation into a single method that does the initialization.

De: Alan McGovern <notifications@github.com>
Responder a: mono/xwt <reply@reply.github.com>
Fecha: dimarts, 5 de setembre de 2017 a les 12:41
Para: mono/xwt <xwt@noreply.github.com>
CC: Lluis Sanchez Gual <llsan@microsoft.com>, Review requested <review_requested@noreply.github.com>
Asunto: Re: [mono/xwt] Handle context menu (#717)


Do you want every place which instantiates a ButtonEventArgs object to explicitly set the correct value for this property?

Xwt.WPF/Xwt.WPFBackend/WidgetBackend.cs
```diff
@@ -648,6 +648,7 @@ void WidgetMouseDownHandler (object o, MouseButtonEventArgs e)
 		void WidgetMouseUpHandler (object o, MouseButtonEventArgs e)
 		{
 			var args = e.ToXwtButtonArgs (Widget);
+			args.ShouldShowContextMenu = e.ChangedButton == MouseButton.Right;
```
alanmcgovern: Can we make `e.ToXwtButtonArgs` set the `ShouldShowContextMenu` property?

slluis: I guess so

matrosov-nikita: @alanmcgovern, @slluis, done.

Xwt.XamMac/Xwt.Mac/Util.cs
```diff
@@ -240,6 +240,21 @@ public static NSBitmapImageFileType ToMacFileType (this ImageFileType type)
...
 		}
 
+		public static bool TriggerContextMenu (this NSEvent theEvent)
+		{
+			if (theEvent.ButtonNumber == 1 &&
+					(NSEvent.CurrentPressedMouseButtons & 1 | NSEvent.CurrentPressedMouseButtons & 4) == 0) {
+				return true;
+			}
+
+			if (theEvent.ButtonNumber == 0 && Keyboard.CurrentModifiers == ModifierKeys.Control &&
```
chkn: Instead of `Keyboard.CurrentModifiers == ModifierKeys.Control` we should probably use `(theEvent.ModifierFlags & NSEventModifierMask.ControlKeyMask) != 0`

Xwt.XamMac/Xwt.Mac/Util.cs
```diff
@@ -240,6 +240,21 @@ public static NSBitmapImageFileType ToMacFileType (this ImageFileType type)
 			}
 		}
 
+		public static bool TriggerContextMenu (this NSEvent theEvent)
```
chkn: Should we call this `TriggersContextMenu` to match the extension method in the Gtk backend?

Xwt/Xwt/ButtonEventArgs.cs
```diff
@@ -80,6 +80,12 @@ public ButtonEventArgs ()
 		/// </summary>
 		/// <value>The number of the presses of the same button.</value>
 		public int MultiplePress { get; set; }
+
+		/// <summary>
+		/// Gets or sets value indicating whether context menu should be shown
+		/// </summary>
+		/// <value><c>true</c> if this context menu should be shown; otherwise, <c>false</c>.</value>
+		public bool ShouldShowContextMenu { get; set; }
```
chkn: IMHO we should call this property something more like `IsContextMenuTrigger`, because it is describing the nature of the event rather than prescribing what the app should do

matrosov-nikita:
@chkn, thanks, fixed.

alanmcgovern:
Thanks guys!

PR #722, opened 2017-09-14
ExTreeView: Fixed missing first root element

### LINKED ISSUE #720: WPF: The first element of TreeView is not displayed
## The issue
If I create a TreeView, bind it to a TreeStore, create a new Node in it and then try to select it, it becomes invisible, while second and all items that follow are rendered just fine.

The code:
```C#
treeStore.AddNode().SetValue(f, "New element");
treeView.SelectRow(navigator.CurrentPosition);
```

Here's a small repro app I made with XWT Libraries built from master:
https://github.com/alsorokin/xwt-wpf-repro

Repro steps:
1. Select the second tab
2. Click "Add item" button
3. See that no item is shown
4. Click "Add item" button again
5. See that the second item is shown

Workaround:
1. Restart the app and click "Add item" button without selecting the second tab
2. Select the second tab
3. See that the first item is shown

## Bonus
I've found that if I comment this line out, things seem to work:
https://github.com/mono/xwt/blob/master/Xwt.WPF/Xwt.WPFBackend/ExTreeView.cs#L201

The downside is, if I understand correctly, that selected item is not properly saved in the internal state.

alsorokin:
Manually updating the widget layout only for root elements, as children are not affected by this bug

This fixes https://bugzilla.xamarin.com/show_bug.cgi?id=58917
This closes #720

### DIFF
--- a/Xwt.WPF/Xwt.WPFBackend/ExTreeView.cs
+++ b/Xwt.WPF/Xwt.WPFBackend/ExTreeView.cs
@@ -191,6 +191,8 @@ namespace Xwt.WPFBackend
 			if (parent != null) {
 				items = parent.Items;
 				g = parent.ItemContainerGenerator;
+			} else {
+				UpdateLayout ();
 			}
 
 			foreach (object item in items) {


alanmcgovern:
This may have broken wpf: https://gist.github.com/alanmcgovern/aebe0fe47a68cfdd1429397b6d8f5d11

alanmcgovern:
Screencast of the crash: https://screencast.com/t/5RTlpt0zRD

alanmcgovern:
the stacktrace has this addition in it: https://screencast.com/t/0KfcFZDDr

PR #726, opened 2017-09-21
D15 5 update

alanmcgovern:


alanmcgovern:
Xwt was branched 2 weeks earlier than the rest of the products so i'm merging master in now. We expected the xwt 15-5 branch to be refreshed when the products branched a few days ago, but that didn't happen. We do depend on some of these changes.

PR #727, opened 2017-09-26
ExTreeView: Don't try to explicitly render virtualized items

alsorokin:
this fixes https://bugzilla.xamarin.com/show_bug.cgi?id=58917


alanmcgovern:
@DavidKarlas @sevoku 

We're reverting back to the original behaviour as it seems to be the most correct!

PR #736, opened 2017-10-05
[Gtk] Optimize TextIndexer

Therzok:


sevoku:
And they pass: https://ci.appveyor.com/project/xwt/xwt/build/0.2.41-doafqgsm/tests

Therzok:
It would be better for it to be public, in case anyone needs indexing between pango and c#

PR #738, opened 2017-10-13
[Xwt] Remember to set `IsReallocating` to false...

alanmcgovern:


alanmcgovern:
That's a bit of an oops, but it's been there for a while so nothing must depend on it really :)

PR #739, opened 2017-10-19
TreeView: Let user drag multiple selected items

alsorokin:
https://bugzilla.xamarin.com/show_bug.cgi?id=60266

### DIFF
--- a/Xwt.Gtk/Xwt.GtkBackend/TableViewBackend.cs
+++ b/Xwt.Gtk/Xwt.GtkBackend/TableViewBackend.cs
@@ -29,6 +29,7 @@ using Xwt.Backends;
 using Gtk;
 using System.Collections.Generic;
 using System.Linq;
+using Gdk;
 #if XWT_GTK3
 using TreeModel = Gtk.ITreeModel;
 #endif
@@ -498,10 +499,14 @@ namespace Xwt.GtkBackend
 	class CustomTreeView: Gtk.TreeView
 	{
 		WidgetBackend backend;
-		
+		TreePath delayedSelection;
+		TreeViewColumn delayedSelectionColumn;
+
 		public CustomTreeView (WidgetBackend b)
 		{
 			backend = b;
+			base.DragBegin += (_, __) =>
+				delayedSelection = null;
 		}
 
 		static CustomTreeView ()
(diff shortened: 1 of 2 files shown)


Xwt.Gtk/Xwt.GtkBackend/TableViewBackend.cs
```diff
@@ -514,6 +519,40 @@ static CustomTreeView ()
...
+				return result;
+			}
+			return base.OnButtonPressEvent (evnt);
+		}
+
+		protected override bool OnButtonReleaseEvent (EventButton evnt)
+		{
+			// Now, if mouse hadn't moved, we are certain that this was just a click. Proceed as usual.
+			if (delayedSelection != null) {
+				if (evnt.X.Equals (delayedSelectionX) && evnt.Y.Equals (delayedSelectionY))
```
chkn: Checking for exact equality seems a bit rigid and we are making assumptions about what constitutes a drag. How about we remove `delayedSelectionX` and `delayedSelectionY` and instead set `delayedSelection` to null in the drag started event?

Xwt.Gtk/Xwt.GtkBackend/TableViewBackend.cs
```diff
@@ -514,6 +519,40 @@ static CustomTreeView ()
...
+			GetPathAtPos ((int)evnt.X, (int)evnt.Y, out treePath, out column);
+			var ctrlShiftMask = (evnt.State & (Gdk.ModifierType.ShiftMask | Gdk.ModifierType.ControlMask | Gdk.ModifierType.Mod2Mask));
+			if (evnt.Button == 1 && this.Selection.PathIsSelected (treePath) && ctrlShiftMask == 0) {
+				delayedSelection = treePath;
+				delayedSelectionColumn = column;
+				delayedSelectionX = evnt.X;
+				delayedSelectionY = evnt.Y;
+				Selection.SelectFunction = (_, __, ___, ____) => false;
+				var result = base.OnButtonPressEvent (evnt);
+				Selection.SelectFunction = (_, __, ___, ____) => true;
```
chkn: I think we should probably restore the previous value for `Selection.SelectFunction` here, in case it had been set to something custom. Also, this should be done in a finally block in case the base handler happens to throw an exception.

Xwt.WPF/Xwt.WPFBackend/ExTreeViewItem.cs
```diff
@@ -194,6 +206,22 @@ protected override void OnGotFocus(RoutedEventArgs e)
 		{
 			BringIntoView();
 			//We can't allow TreeViewItem(our base class) to get this message(OnGotFocus) because it will also select this item which we don't want
-		}
+		}
+
+		private bool ShiftPressed
```
chkn: You added this property but don't seem to use it anywhere?

Xwt.Gtk/Xwt.GtkBackend/TableViewBackend.cs
alsorokin: We can't save previous `SelectFunction` value here because this property has no getter 😫

Xwt.Gtk/Xwt.GtkBackend/TableViewBackend.cs
alsorokin: The problem with this is that I can't seem to find propert drag started or drag motion event. I've tried overriding
```
protected override bool OnDragMotion(Gdk.DragContext context, int x, int y, uint time)
{
	return true;
}
```
but it doesn't get called. 😓

alsorokin:
Hey @chkn thanks for your comments. I've fixed what I could, and replied to what can't seem to be fixed.

Xwt.Gtk/Xwt.GtkBackend/TableViewBackend.cs
alsorokin: Nevermind, I've found a way :)

Xwt.Gtk/Xwt.GtkBackend/TableViewBackend.cs
```diff
@@ -498,10 +499,14 @@ public void QueueDraw (object target, Gtk.TreeIter iter)
...
 	{
 		WidgetBackend backend;
-		
+		TreePath delayedSelection;
+		TreeViewColumn delayedSelectionColumn;
+
 		public CustomTreeView (WidgetBackend b)
 		{
 			backend = b;
+			base.DragBegin += (_, __) =>
```
sevoku: can't we use an `OnDragBegin` override for this?

sevoku:
I wonder whether the upstream behaviour is intended (on Linux?) or broken?

Xwt.Gtk/Xwt.GtkBackend/TableViewBackend.cs
alsorokin: Tried that, it doesn't get call for some reason :(

alsorokin:
I can't fathom this behavior being intended on any platform 😐

alsorokin:
I saw some people on the Internets struggling with this gtk behavior too:
http://python.6.x6.nabble.com/gtk-TreeView-Multiple-Selection-Drag-and-Drop-td1947055.html

Xwt.Gtk/Xwt.GtkBackend/TableViewBackend.cs
chkn: Ah I see. Well it's probably ok because xwt owns the `CustomTreeView` anyway.

Xwt.Gtk/Xwt.GtkBackend/TableViewBackend.cs
sevoku: Hhm, that's strange. Usually some handlers need to be attributed with `[GLib.ConnectBefore]`, but overrides should just work. But in this case it's ok to use the event then.

sevoku:
So, the GTK impl. looks good, but I can't approve the WPF part, because I have no working Windows machine right now. @chkn or @Therzok: could you verify the WPF part?

alsorokin:
Squashed, rebased.

PR #743, opened 2017-10-23
D15 5 backports

alanmcgovern:


alanmcgovern:
It looks like the last batch of backports were all cherry-picked instead of merged. That's making the commits show up in the list, but if you look at the actual code diff you'll see it's just the two small changes we need (the one to properly reset the bool and the one to ensure we measure when measure is invalid).

PR #744, opened 2017-10-23
[WPF] CustomScrollViewPort: Ensure content offset is updated if child size changes

chkn:
If a larger child is replaced by a smaller one, we need to ensure the content offset
is re-clamped based on the new extents, otherwise the child may be scrolled entirely
out of the viewport.

### DIFF
--- a/Xwt.WPF/Xwt.WPFBackend/CustomScrollViewPort.cs
+++ b/Xwt.WPF/Xwt.WPFBackend/CustomScrollViewPort.cs
@@ -196,37 +196,9 @@ namespace Xwt.WPFBackend
 			set;
 		}
 
-		public void SetHorizontalOffset (double offset)
-		{
-			if (offset < 0 || this.viewport.Width >= this.extent.Width)
-				offset = 0;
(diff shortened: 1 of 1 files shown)


Xwt.WPF/Xwt.WPFBackend/CustomScrollViewPort.cs
```diff
@@ -349,6 +355,10 @@ protected override System.Windows.Size ArrangeOverride (System.Windows.Size fina
 					this.viewport = finalSize;
 					ScrollOwner.InvalidateScrollInfo ();
 				}
+
+				// It's possible our child is now smaller than our viewport, in which case we need to
+				//  update the content offset, otherwise the child may be scrolled entirely out of view.
+				SetOffset ();
```
alanmcgovern: 💯

alanmcgovern: 🔨

alanmcgovern:
Can we get this one backported too!

PR #747, opened 2017-10-31
[A11y] Allow native accessible children to be added to Xwt.Accessible

iainx:
Sometimes widgets have subregions that need their own accessible object. This
change allows native accessible objects to be added to Xwt.Accessible objects
as children

### DIFF
--- a/Xwt.Gtk.Mac/GtkMacAccessibleBackend.cs
+++ b/Xwt.Gtk.Mac/GtkMacAccessibleBackend.cs
@@ -24,10 +24,11 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 using System;
-
+using System.Linq;
 using AppKit;
 using Foundation;
 using ObjCRuntime;
+using Xwt.Accessibility;
 using Xwt.GtkBackend;
 
 namespace Xwt.Gtk.Mac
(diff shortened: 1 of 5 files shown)


Xwt.Gtk.Mac/GtkMacAccessibleBackend.cs
```diff
@@ -83,5 +84,30 @@ internal static INSAccessibility GetNSAccessibilityElement (GLib.Object o)
...
+			}
+		}
+
+		public override void RemoveChild (object nativeChild)
+		{
+			var nsa = GetNSAccessibilityElement (widget.Accessible);
+			var parent = nsa as NSAccessibilityElement;
+
+			if (parent != null) {
+				var children = parent.AccessibilityChildren.ToList ();
```
Therzok: Can we not create a new array that of `parent.AccessibilityChildren.Length - 1` and skip `nativeChild`?

sevoku: What would happen if `nativeChild` is not in `parent.AccessibilityChildren`?
But maybe we can skip 1 iteration, if we remove `children.Remove ((NSObject)nativeChild);` and just enumerate to an array like:
`parent.AccessibilityChildren = parent.AccessibilityChildren.Where(c => c != nativeChild).ToArray ();`

Xwt.Gtk/Xwt.GtkBackend/AccessibleBackend.cs
```diff
@@ -157,6 +157,18 @@ public void InitializeBackend (object frontend, ApplicationContext context)
 			}
 		}
 
+		public virtual void AddChild (object nativeChild)
+		{
+			// TODO
+			throw new NotImplementedException ();
```
sevoku: It's better to do nothing here, otherwise we'll need to catch this exception everywhere. Just keep the TODO comment.

Xwt.XamMac/Xwt.Mac/AccessibleBackend.cs
```diff
@@ -173,6 +173,18 @@ bool PerformPress ()
 			}
 		}
 
+		public void AddChild (object nativeChild)
+		{
+			// TODO
+			throw new NotImplementedException ();
```
sevoku: Don't we need the same impl like in `GtkMacAccessibleBackend` here?

Xwt.Gtk.Mac/GtkMacAccessibleBackend.cs
Therzok: Good point. That works for me.

Issue #748, opened 2017-11-03
TreeView: Multiple selection still not working properly


alsorokin:
Repro steps:

1. In sample xwt apps, modify "TreeView" sample so that it allows multiple selection:
```C#
public TreeViews ()
{
	...
	view.SelectionMode = SelectionMode.Multiple;
	...
}
```
2. Launch `WpfTest` project
3. Open `Widgets/TreeView` sample
4. Add some child items to any item. They all should be of same depth.
5. Click on the first of the child items
6. Shift-click on the last one

Expected result:
All child items are selected

Actual result:
Only one child item is selected

PR #749, opened 2017-11-03
[WPF] Fixed TreeView multiple child selection not working properly

alsorokin:
This fixes https://github.com/mono/xwt/issues/748

Removed malfunctioning call to `GetTreeItem()` method. It was calling `ItemContainerGenerator.ContainerFromItem()`, which for some reason returns null for any child item.

So while the root of the problem may lie somewhere deeper in the depths of WPF, these changes do fix the issue for us.


alsorokin:
So, while the issue is fixed simply by removing the method call (which I added myself in my previous fix), I also took it upon myself to optimize item selection logic.
Previously, if `view.SelectionMode` was `Single`, then we would select all the items between `shiftStart` and `shiftEnd` one by one anyway (ending up selecting only the item clicked by the user). Now we select only the item user clicked by the user.

PR #754, opened 2017-11-12
Add default button support

sevoku:
This patch adds `Button.IsDefault` support. Depending on the toolkit, default buttons will be activated on special default actions inside a window (usually when the user presses the `Enter` key on the keyboard) and have a different design.

TODO: WPF support


Xwt.WPF/Xwt.WPFBackend/ButtonBackend.cs
```diff
@@ -135,6 +135,8 @@ public Xwt.Drawing.Color LabelColor
 			set { Button.Foreground = ResPool.GetSolidBrush (value.ToWpfColor()); }
 		}
 
+		public bool IsDefault { get; set; } //TODO: Default Button
```
Therzok: I think you just need to route into `Button.Default`

PR #756, opened 2017-11-16
Add TreeView multi-selection toggle to samples

chkn:
This helps testing changes like https://github.com/mono/xwt/pull/739

### DIFF
--- a/TestApps/Samples/Samples/TreeViews.cs
+++ b/TestApps/Samples/Samples/TreeViews.cs
@@ -168,6 +168,16 @@ namespace Samples
 				Console.WriteLine("Collapsed: " + val);
 			};
 
+			RadioButtonGroup group = new RadioButtonGroup ();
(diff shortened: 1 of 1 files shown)


TestApps/Samples/Samples/TreeViews.cs
```diff
@@ -168,6 +168,16 @@ public TreeViews ()
 				Console.WriteLine("Collapsed: " + val);
 			};
 
+			RadioButtonGroup group = new RadioButtonGroup ();
```
Therzok: Don't we need to set an active one?

sevoku: No, the toolkits usually always activate the first option in a group by default.

sevoku:
Hhm, interesting, this reveals some issues when deleting multiple rows. `SelectedRows` could change while interating, and child rows could become invalid, if a parent is deleted. We need to investigate that.

Issue #757, opened 2017-11-23
More cursor types


paymicro:
It would be nice to have some more cursor types available. We would like to use them in Xamarin iOS Designer.
We're lacking the diagonal resize, drag and openhand cursors.

PR #758, opened 2017-11-23
More cursor types

paymicro:
Add more cursor types

Fixes https://github.com/mono/xwt/issues/757

### DIFF
--- a/Xwt.Gtk/Xwt.GtkBackend/WidgetBackend.cs
+++ b/Xwt.Gtk/Xwt.GtkBackend/WidgetBackend.cs
@@ -192,6 +192,8 @@ namespace Xwt.GtkBackend
 					ctype = Gdk.CursorType.Crosshair;
 				else if (cursor == CursorType.Hand)
 					ctype = Gdk.CursorType.Hand1;
+				else if (cursor == CursorType.Hand2 || cursor == CursorType.DragCopy)
+					ctype = Gdk.CursorType.Hand2;
 				else if (cursor == CursorType.IBeam)
 					ctype = Gdk.CursorType.Xterm;
 				else if (cursor == CursorType.ResizeDown)
@@ -206,6 +208,16 @@ namespace Xwt.GtkBackend
 					ctype = Gdk.CursorType.SbHDoubleArrow;
 				else if (cursor == CursorType.ResizeUpDown)
 					ctype = Gdk.CursorType.SbVDoubleArrow;
+				else if (cursor == CursorType.ResizeNE)
+					ctype = Gdk.CursorType.TopRightCorner;
+				else if (cursor == CursorType.ResizeNW)
+					ctype = Gdk.CursorType.TopLeftCorner;
+				else if (cursor == CursorType.ResizeSE)
+					ctype = Gdk.CursorType.BottomRightCorner;
+				else if (cursor == CursorType.ResizeSW)
+					ctype = Gdk.CursorType.BottomLeftCorner;
+				else if (cursor == CursorType.NotAllowed)
+					ctype = Gdk.CursorType.XCursor;
 				else if (cursor == CursorType.Move)
 					ctype = Gdk.CursorType.Fleur;
 				else if (cursor == CursorType.Wait)
--- a/Xwt.XamMac/Xwt.Mac/ViewBackend.cs
+++ b/Xwt.XamMac/Xwt.Mac/ViewBackend.cs
@@ -222,6 +222,8 @@ namespace Xwt.Mac
 				Cursor = NSCursor.CrosshairCursor;
 			else if (cursor == CursorType.Hand)
 				Cursor = NSCursor.OpenHandCursor;
+			else if (cursor == CursorType.Hand2)
+				Cursor = NSCursor.PointingHandCursor;
 			else if (cursor == CursorType.IBeam)
 				Cursor = NSCursor.IBeamCursor;
 			else if (cursor == CursorType.ResizeDown)
@@ -241,6 +243,10 @@ namespace Xwt.Mac
 				Cursor = NSCursor.ArrowCursor;
 			else if (cursor == CursorType.Move)
 				Cursor = NSCursor.ClosedHandCursor;
+			else if (cursor == CursorType.DragCopy)
+				Cursor = NSCursor.DragCopyCursor;
+			else if (cursor == CursorType.NotAllowed)
+				Cursor = NSCursor.OperationNotAllowedCursor;
 			else
 				Cursor = NSCursor.ArrowCursor;
 		}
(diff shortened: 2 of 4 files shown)


Xwt/Xwt/CursorType.cs
```diff
@@ -53,13 +53,18 @@ internal CursorType (string id)
 		public static readonly CursorType ResizeUp = new CursorType ("ResizeUp");
 		public static readonly CursorType ResizeDown = new CursorType ("ResizeDown");
 		public static readonly CursorType ResizeUpDown = new CursorType ("ResizeUpDown");
+		public static readonly CursorType Resize45 = new CursorType("Resize45");
+		public static readonly CursorType Resize135 = new CursorType("Resize135");
+		public static readonly CursorType Resize225 = new CursorType("Resize225");
+		public static readonly CursorType Resize315 = new CursorType("Resize315");
```
sevoku: I think WPF style naming would be better (like `NESW`)

Xwt.Gtk/Xwt.GtkBackend/WidgetBackend.cs
```diff
@@ -185,38 +185,23 @@ public void SetCursor (CursorType cursor)
...
-				else if (cursor == CursorType.Wait)
-					ctype = Gdk.CursorType.Watch;
-				else if (cursor == CursorType.Help)
-					ctype = Gdk.CursorType.QuestionArrow;
-				else if (cursor == CursorType.Invisible)
-					ctype = (Gdk.CursorType)(-2); // Gdk.CursorType.None, since Gtk 2.16
-				else
-					ctype = Gdk.CursorType.Arrow;
-				
+				var cursors = new Dictionary<CursorType, Gdk.CursorType> {
```
sevoku: Why should we use an expensive `Dictionary` here? `if/else` or a switch is much more efficient.

Xwt/Xwt/CursorType.cs
```diff
@@ -53,13 +53,17 @@ internal CursorType (string id)
 		public static readonly CursorType ResizeUp = new CursorType ("ResizeUp");
 		public static readonly CursorType ResizeDown = new CursorType ("ResizeDown");
 		public static readonly CursorType ResizeUpDown = new CursorType ("ResizeUpDown");
+		public static readonly CursorType ResizeNESW = new CursorType("ResizeNESW");
+		public static readonly CursorType ResizeNWSE = new CursorType("ResizeNWSE");
 		public static readonly CursorType Hand = new CursorType ("Hand");
+		public static readonly CursorType OpenHand = new CursorType("OpenHand");
```
sevoku: We should rethink the naming here. We map `Hand` to `NSCursor.OpenHandCursor` on Mac already, so the naming is confusing and we can't simply swap them (could break Xwt users who expect the open hand cursor for `Hand`).

We should deprecate `Hand` (with `[Obsolete("Please use Hand1 or Hand2 instead")]`) and either use the Gtk style with `Hand1` and `Hand2`, or use `ClosedHand`, `OpenHand`. I'd prefer the first variant, cause we have only one hand cursor on Windows and the Gtk cursors are not necessarily `open`/`closed`, they're just two different hand styles.

Also the Cocoa `NSCursor.ClosedHandCursor` is mapped to `CursorType.Move` which makes sense. There is a third `NSCursor.PointingHandCursor` (not used right now) which would perfectly fit `CursorType.Hand2`.

|CursorType|Gtk|Cocoa|Wpf|
|---|---|---|---|
|`CursorType.Hand1`|`Gdk.CursorType.Hand1`|`NSCursor.OpenHandCursor`|`Cursors.Hand`|
|`CursorType.Hand2`|`Gdk.CursorType.Hand2`|`NSCursor.PointingHandCursor`|`Cursors.Hand`|

Xwt/Xwt/CursorType.cs
```diff
@@ -53,13 +53,17 @@ internal CursorType (string id)
...
 		public static readonly CursorType Move = new CursorType ("Move");
 		public static readonly CursorType Wait = new CursorType ("Watch");
 		public static readonly CursorType Help = new CursorType ("Help");
 		public static readonly CursorType Invisible = new CursorType ("Invisible");
+		public static readonly CursorType Drag = new CursorType("Drag");
```
sevoku: Could you also add Gtk (on Linux it's always mapped to `Gdk.CursorType.Hand2`)/Wpf (?) mappings for `CursorType.Drag`?

Xwt.XamMac/Xwt.Mac/ViewBackend.cs
```diff
@@ -214,6 +214,19 @@ public void NotifyPreferredSizeChanged ()
...
+
+		static NSCursor GetCursorFromResource(string resourceName, CGPoint center) {
+			var resource = Drawing.Image.FromResource(resourceName);
+			if (resource == null)
+				return NSCursor.ArrowCursor;
+			var image = (NSImage)macToolkit?.GetNativeImage(resource);
+			return image != null ? new NSCursor(image, center) : NSCursor.ArrowCursor;
+		}
+
+		static readonly NSCursor resizeNWSE = GetCursorFromResource("resize-NWSE-16.png", new CGPoint(8, 8));
```
sevoku: We need Retina support here with 16@2x, otherwise the cursors appear blurry.

paymicro:
Preview of the diagonal icons in Mac : https://www.screencast.com/t/XhCUpAta4i

sevoku:
@paymicro I see you working on the additional Cocoa cursors, which is really nice, but it will be a bigger task and needs some investigation.

1. Where did you get the new images from? We must ensure that we don't violate any licenses.
2. I think there might be a more elegant way to use original mac OS resize icons, see https://stackoverflow.com/questions/10733228/native-osx-lion-resize-cursor-for-custom-nswindow-or-nsview or https://stackoverflow.com/questions/27242353/cocoa-predefined-resize-mouse-cursor. We should investigate that approach and go with native if possible.

I'd propose to split the PRs, to unblock the initial proposed API change. This PR would add the new API + impls without the new custom Cocoa cursors, and an other one (based on this) the Cocoa diagonal resize cursors.

sevoku:
Ok, so it's under Apples license which is definitely not compatible with our MIT. We definitely can't use those icons (even if we rotate them). Lets split the PRs and discuss it in the new PR.

paymicro:
@sevoku are there any outstanding issues with this PR or can it be merged?

Xwt.XamMac/Xwt.Mac/ViewBackend.cs
```diff
@@ -241,6 +243,8 @@ public void SetCursor (CursorType cursor)
 				Cursor = NSCursor.ArrowCursor;
 			else if (cursor == CursorType.Move)
 				Cursor = NSCursor.ClosedHandCursor;
+			else if (cursor == CursorType.Drag)
```
sevoku: I've double checked the default Finder behaviour on Mac and `NSCursor.DragCopyCursor` seems to be wrong. The default drag operation has no special cursor (defaults to `NSCursor.ArrowCursor`), only if you press the Alt key it will change to the `NSCursor.DragCopyCursor`.

paymicro: I renamed the cursor type `Drag` to `DragCopy`.
Also added `NotAllowed` cursor type.

PR #760, opened 2017-11-29
Leak fix attempt

Therzok:
Doesn't fix NativeWindowFrame and (possible?) CellView leaks

### DIFF
--- a/Xwt/Xwt.Backends/BackendHost.cs
+++ b/Xwt/Xwt.Backends/BackendHost.cs
@@ -59,7 +59,7 @@ namespace Xwt.Backends
 	/// <summary>
 	/// The BackendHost is the link between an Xwt frontend and a toolkit backend.
 	/// </summary>
-	public class BackendHost: EventHost
+	public class BackendHost: EventHost, IDisposable
 	{
 		IBackend backend;
 		bool usingCustomBackend;
@@ -131,8 +131,10 @@ namespace Xwt.Backends
 		/// </summary>
 		protected virtual void OnBackendCreated ()
 		{
-			foreach (var ev in DefaultEnabledEvents)
+			foreach (var ev in DefaultEnabledEvents) {
+				enabledEvents.Add (ev);
 				Backend.EnableEvent (ev);
+			}
 		}
 		
 		/// <summary>
@@ -171,12 +173,15 @@ namespace Xwt.Backends
 			}
 		}
 
+		List<object> enabledEvents = new List<object> ();
+
 		/// <summary>
 		/// Enables an event with the specified identifier.
 		/// </summary>
 		/// <param name="eventId">Event identifier (must be a valid event enum value).</param>
 		protected override void OnEnableEvent (object eventId)
 		{
+			enabledEvents.Add (eventId);
 			Backend.EnableEvent (eventId);
 		}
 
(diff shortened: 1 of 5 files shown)


Xwt/Xwt.Backends/BackendHost.cs
```diff
@@ -171,12 +173,15 @@ protected void LoadBackend ()
 			}
 		}
 
+		List<object> enabledEvents = new List<object> ();
```
sevoku: Wouldn't it better to use `HashSet`?

Therzok: Could be.

Xwt/Xwt/WindowFrame.cs
```diff
@@ -74,6 +74,7 @@ public class WindowFrame: XwtComponent, IAnimatable
 		bool pendingReallocation;
 		Image icon;
 		WindowFrame transientFor;
+		protected bool owned;
```
sevoku: this should be `true` by default and only be false for `NativeWindowFrame`

Xwt/Xwt/WindowFrame.cs
```diff
@@ -83,7 +84,7 @@ protected override void OnBackendCreated ()
 				base.OnBackendCreated ();
 				Parent.location = Backend.Bounds.Location;
 				Parent.size = Backend.Bounds.Size;
-				Backend.EnableEvent (WindowFrameEvent.BoundsChanged);
+				OnEnableEvent (WindowFrameEvent.BoundsChanged);
```
sevoku: what if we do this only if `owned == true`? Enabling this by default is only required by the initial window pisition/size logic afair (@slluis? correct me if I'm wrong). If the client needs all the bounds/sizing feature, it should enable/disable this event manually.

Xwt/Xwt/WindowFrame.cs
Therzok: Yes, sorry, I screwed up.

Xwt/Xwt.Backends/BackendHost.cs
slluis: Actually, since we don't have to check if events are registered, a list is better.

Xwt/Xwt/WindowFrame.cs
slluis: We also use that event to keep the Bounds property up to date, so I guess we still need it.

sevoku: Hhm, right, but in that case we could query `Backend.Bounds` directly. And since we don't need `Reallocate ()` if the window is embedded this should do the trick I think:

```diff
--- a/Xwt/Xwt/WindowFrame.cs
+++ b/Xwt/Xwt/WindowFrame.cs
@@ -84,7 +84,8 @@ namespace Xwt
 				base.OnBackendCreated ();
 				Parent.location = Backend.Bounds.Location;
 				Parent.size = Backend.Bounds.Size;
-				OnEnableEvent (WindowFrameEvent.BoundsChanged);
+				if (Parent.owned)
+					OnEnableEvent (WindowFrameEvent.BoundsChanged);
 			}
 			
 			public void OnBoundsChanged (Rectangle bounds)
@@ -404,7 +405,9 @@ namespace Xwt
 		internal virtual Rectangle BackendBounds {
 			get {
 				BackendHost.EnsureBackendLoaded ();
-				return new Rectangle (location, size);
+				if (owned)
+					return new Rectangle (location, size);
+				return Backend.Bounds;
 			}
 			set {
 				size = value.Size;

```

Therzok:
Pushed fixups!

PR #761, opened 2017-11-29
Bump TargetFrameworkVersion to 4.5

sevoku:


elsupergomez:
This is similar to https://github.com/mono/xwt/pull/550

PR #767, opened 2018-01-05
Return null if unable to find an image

KirillOsenkov:
Avoids a first-chance exception. We catch an InvalidOperationException later anyway, so might as well just use null to represent a missing value.

### DIFF
--- a/Xwt/Xwt.Backends/ImageBackendHandler.cs
+++ b/Xwt/Xwt.Backends/ImageBackendHandler.cs
@@ -43,7 +43,7 @@ namespace Xwt.Backends
 		{
 			using (var s = asm.GetManifestResourceStream (name)) {
 				if (s == null)
-					throw new InvalidOperationException ("Resource not found: " + name);
+					return null;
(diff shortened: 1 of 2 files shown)


KirillOsenkov:
A corresponding change in MonoDevelop will need to be made.

sevoku:
Yeah, I agree with @Therzok, but a new `bool TryLoadImage(string resource, out Image image)` might be useful in some cases, maybe we should add that instead?

KirillOsenkov:
My concern is here:
http://source.monodevelop.com/#WindowsPlatform/ImageHelper.cs,46

We know we won't find the .png, so we just catch the InvalidOperationException, ignore the information anyway and attempt to call GetIcon.

All I really want is to get rid of that first-chance exception because it's super noisy. It fails for about 15-20 images during startup path, and more when you open menus.

Any suggestions on the best way to fix this?

KirillOsenkov:
I've found a better fix that is WindowsPlatform-only and doesn't need changes to Xwt. Abandoning this one.

PR #772, opened 2018-02-03
Fix ITreeDataSource implementation

rodrmoya:


rodrmoya:
This works now, but see https://github.com/mono/xwt/pull/772/commits/5c8faca3247b22fe4313a025c99944e2a95028e7

Xwt.Gtk/Xwt.GtkBackend/CustomTreeModel.cs
```diff
@@ -178,8 +175,6 @@ public bool IterNext (ref Gtk.TreeIter iter)
 			TreePosition pos = NodeFromIter (iter);
 			TreePosition parent = source.GetParent (pos);
 			int i = GetIndex (parent, pos);
-			if (source.GetChildrenCount (parent) >= i)
-				return false;
```
Therzok: Why remove this validation?

rodrmoya: It's just overhead, see the lines after that, which just call the data source's .GetChild for the next sibling, which would return null if there is no child at that position, so there's no need to call this.

Also, AFAICS, the condition is wrong, as it checks if there are more children than the current position,while it should check if it has *less* to return `false` (i.e. no next iter).

rodrmoya:
Better fixed at https://github.com/mono/xwt/pull/773 so closing

PR #773, opened 2018-02-05
Fixed ITreeDataSource support in Gtk.TreeView

slluis:
The fix implements caching of indexes, to avoid scanning child lists every
time an index is required.

Fixed Gtk.TreeViewBacked, which didn't support custom TreePosition
implementations.

Added new sample.


sevoku:
Hhm, this broke both CIs

PR #779, opened 2018-02-13
[Gtk] Fix editable cells

sevoku:
The DragDrop fix introduced in 2107af1ab793598774cbae77afd44d76e2943a18
breaks cell editing by delaying row selection and reselecting a row on
the ButtonReleaseEvent stops the current editing operation immediatly.

To allow cell editing, we need to check whether multiple selection is enabled
and that already more than one row has been selected, before we do the
delayed selection.

Fixes VSTS #564512


alsorokin:
Cool, the code looks good to me 👍

PR #782, opened 2018-02-16
Add BorderVisible property to TreeView

rodrmoya:


rodrmoya:
Yes, saw it, so changed now

PR #783, opened 2018-02-19
[WPF] Properly fire "Button Pressed" and "Button Released" events for…

alsorokin:
… TreeView

This fixes the discrepancy between Gtk and Wpf implementations.

In Gtk, if you click any of the TreeView's rows, `TreeView.ButtonPressed` event is fired. On WPF, only `TreeViewItem.ButtonPressed` is fired.


Therzok:
I'd rephrase the commit message to something like:
`Fire ButtonPressed and ButtonReleased events for WPF, similar to the Gtk backend`

alsorokin:
@sevoku makes sense, thanks for the review. Updated the commit.
@Therzok thanks, I've changed the commit name.

PR #786, opened 2018-02-28
Misc tree view fixes

slluis:
* Added support for disabling selection in TreeView
* Fixed cell position calculation issues
* Added missing conversion to pango units when getting index from coordinates. Also fixed out of array bounds exception.
* Fixed several mouse handling issues in TreeView
* Added QueueResize method to CanvasCellView
* New TreeView sample


sevoku:
Besides my previous comment, it works! But we need Cocoa support for this, it doesn't size correctly.

sevoku:
Uff.. no Gtk3 is still broken, I'll take a look...

PR #789, opened 2018-03-03
[Gtk] Fix CellBounds calculation

sevoku:


Xwt.Gtk/Xwt.GtkBackend.CellViews/CustomCellRenderer.cs
```diff
@@ -57,49 +58,62 @@ internal void StartDrawing (Rectangle ba, Rectangle ca, Gtk.CellRendererState fl
 			hasFocus = (flags & CellRendererState.Focused) != 0;
 			isPrelit = (flags & CellRendererState.Prelit) != 0;
 
-			isDrawing = true;
+			// Gtk will rerender the cell on every status change, hence we can expect the values
+			// set here to be valid until the geometry or any other status of the cell and tree changes.
+			// Setting shown=true ensures that those values are always reused instead if queried
+			// from the parent tree after the cell has been rendered.
```
slluis: Let's say you have a tree with 10 rows. After rendering the 10 rows, the data cached in the renderer is from the 10th row. Now if I move the mouse over the first row, it will be using the cellArea from the 10th row. Moving the mouse over a cell doesn't change its status, and the tree has no need to render it.

Xwt.Gtk/Xwt.GtkBackend.CellViews/CustomCellRenderer.cs
```diff
@@ -57,49 +58,62 @@ internal void StartDrawing (Rectangle ba, Rectangle ca, Gtk.CellRendererState fl
...
+			shown = true;
+		}
+
+		protected override void OnLoadData ()
+		{
+			if (!CurrentIter.Equals (lastIter)) {
```
slluis: You can't use Equals to compare TreeIters. You have to get the TreePath from both iters and compare them.

Xwt.Gtk/Xwt.GtkBackend.CellViews/CustomCellRenderer.cs
sevoku: That's the point, it WILL redraw, I don't know the exact logic responsible for this, but even if you disable all the events the TreeView will invalidate an area behind the mouse pointer and redraw the renderes inside that area.

Xwt.Gtk/Xwt.GtkBackend.CellViews/CustomCellRenderer.cs
sevoku: Hhm, we're only interested in whether the iter has changed or not, if LoadData has been called with a different iter, the new and the old one will have different UserData, which is actually chacked by `Equals`: https://github.com/mono/gtk-sharp/blob/gtk-sharp-2-12-branch/gtk/TreeIter.custom#L33, and it works exactly as expected.

Xwt.Gtk/Xwt.GtkBackend.CellViews/CustomCellRenderer.cs
sevoku: Actually it will call the data func, which will Load the cell data and that will result in a redraw.

Xwt.Gtk/Xwt.GtkBackend.CellViews/CustomCellRenderer.cs
slluis: Maybe in this case it doesn't matter, but it is possible that two iters with different UserData are pointing to the same row. That is, `TreeIter.Equals()` tells you if two iters are exactly the same, not if they point to the same row.

Xwt.Gtk/Xwt.GtkBackend.CellViews/CustomCellRenderer.cs
slluis: OK then.

Xwt.Gtk/Xwt.GtkBackend.CellViews/CustomCellRenderer.cs
sevoku: Yeah, but it doesn't really matter in this case, because I don't think that there will be ever a Iter pair with same content pointing to different rows. And false positives (different iters but row unchanged) should be rare, I even think that in that case we'd need to invalidate the props too, because the model would have been changed.
I also wanted to avoid adding all the TreePath conversion stuff, we'd need to keep a ref to the parent Tree in the CustomCellRenderer or add an additional func to ICellRendererTarget for that purpose.

Issue #790, opened 2018-03-05
TestApps/Samples/Samples/TreeViewEvents.cs


elsupergomez:
The example use code convention supported only in C# 7 in lines 105, 205 and 232.
I use mono 4.8 (last package available for Fedora and CentOS) that not has full support for C#7 but has for C# 6.

I suggest use C# 6 standard.
Thanks

sevoku:
Oh, yes, that needs to be fixed

elsupergomez:
Thanks.
It work ok now.

PR #791, opened 2018-03-05
Fix build with mono 4.8

sevoku:
fix issue #790


sevoku:
@slluis I think we should set `<LangVersion>6</LangVersion>` in all projects. It's not only AppVeyor which doesn't support C# 7, but also stable mono on various Linux distributions. We should keep it compatible as long as we don't really rely on C#7 features.

sevoku:
see #792

PR #794, opened 2018-03-06
Fixes issue rendering when node is collapsed and text has new lines

netonjm:
Fixes issue rendering when node is collapsed and text has new lines
It happens when text doesn't exceed the maxwidth and we have
multiple lines because the new lines


TestApps/Samples/Samples/TreeViewEvents.cs
```diff
@@ -48,7 +48,7 @@ public TreeViewEvents ()
 			tree.Columns.Add (col);
 
 			var node = store.AddNode ().SetValue (nodeField, new TextNode ("Root 1"));
-			var child = node.AddChild ().SetValue (nodeField, new TextNode ("Very long text. Very long text. Very long text. Very long text. Very long text. Very long text. Very long text. Very long text."));
+			var child = node.AddChild ().SetValue (nodeField, new TextNode ($"Very long text with NewLines. {Environment.NewLine} Very long text with NewLines.{Environment.NewLine} Very long text. Very long text. Very long text. Very long text."));
```
slluis: Just use "\n", no need to embed a call to Environment.NewLine.

TestApps/Samples/Samples/TreeViewEvents.cs
```diff
@@ -125,9 +136,15 @@ protected override Size OnGetRequiredSize ()
 				textSize = layout.GetSize ();
 			}
 
-			status.LastCalculatedHeight = textSize.Height;
+			//in cases when there are multiple lines and they don't execeed the max width we need to care height include the expand
+			double height = textSize.Height;
+			if (!status.Expanded && textSize.Height > defaultHeight) {
```
slluis: This can be simplified. If there are multiple lines, just measure the height of the first line.

TestApps/Samples/Samples/TreeViewEvents.cs
```diff
@@ -100,6 +100,17 @@ class ViewStatus
...
 
+		//defines the default height size based in the current font
+		//if will work if you use same font size in every row,
+		//if not, you'll need check in every OnGetRequiredSize
+		double defaultHeight;
+
+		public ExpandableTextCellView () 
+		{
+			var defaultLayout = new TextLayout() { Text = "A" };
+			defaultHeight = defaultLayout.GetSize().Height;
```
sevoku: This won't work correctly. Because it doesn't take all possible glyphs into account. For example the height of `A` is smaller than the height of `Ag`. It's not possible to measure the final height at this point.
Better measure the first line, as proposed here: https://github.com/mono/xwt/pull/794/files#r172651546

TestApps/Samples/Samples/TreeViewEvents.cs
```diff
@@ -147,9 +164,16 @@ protected override void OnDraw (Context ctx, Rectangle cellArea)
 				layout.SetBackground (Colors.LightBlue, Math.Min (selectionStart, selectionEnd), Math.Abs (selectionEnd - selectionStart));
 
 			// Text doesn't fit. We need to render the expand icon
+			if (textSize.Width > cellArea.Width || textSize.Height > cellArea.Height) {
+
+				if (textSize.Width > cellArea.Width || textSize.Height > cellArea.Height) {
```
sevoku: Why this check again?

netonjm: ups!

Therzok:
What should we do about this PR?

PR #797, opened 2018-03-07
Constrained CanvasCellView sizing

sevoku:
This adds a new `CanvasCellView.GetRequiredSize (SizeConstraint widthConstraint)` func for size calculation and deprecates `CanvasCellView.GetRequiredSize ()`.
The `widthConstraint` will be set to the available width inside the column and row, or be unconstrained to calculate a default desired size.

TODO: WPF support (still uses deprecated calc without constraint).

### DIFF
--- a/Xwt/Xwt.Backends/ICanvasCellViewFrontend.cs
+++ b/Xwt/Xwt.Backends/ICanvasCellViewFrontend.cs
@@ -33,7 +33,9 @@ namespace Xwt.Backends
 		ApplicationContext ApplicationContext { get; }
 		void Draw (object ctxBackend, Rectangle cellArea);
 		Rectangle GetDrawingAreaForBounds (Rectangle cellBounds);
+		[Obsolete("Use OnGetRequiredSize (SizeConstraint widthConstraint)")]
 		Size GetRequiredSize ();
+		Size GetRequiredSize (SizeConstraint widthConstraint);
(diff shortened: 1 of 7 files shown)


Xwt/Xwt.Backends/ICanvasCellViewFrontend.cs
```diff
@@ -33,7 +33,9 @@ public interface ICanvasCellViewFrontend: ICellViewFrontend
 		ApplicationContext ApplicationContext { get; }
 		void Draw (object ctxBackend, Rectangle cellArea);
 		Rectangle GetDrawingAreaForBounds (Rectangle cellBounds);
+		[Obsolete("Use OnGetRequiredSize (SizeConstraint widthConstraint)")]
```
Therzok: There is no `OnGetRequiredSize`

Therzok: Maybe also decrease verbosity by removing the parameter name.

Therzok:
Other than typo in the obsoletion message, LGTM.

Xwt/Xwt.Backends/ICanvasCellViewFrontend.cs
sevoku: yep, thanks!

PR #799, opened 2018-03-09
[Mac] Improve table row height calculation

sevoku:
This patch reduces the amount of required recalculations while resizing rows and fixes row reallocation when the NSTableView is being resized.

### DIFF
--- a/Xwt.XamMac/Xwt.Mac.CellViews/CompositeCell.cs
+++ b/Xwt.XamMac/Xwt.Mac.CellViews/CompositeCell.cs
@@ -88,9 +88,23 @@ namespace Xwt.Mac
 			source.SetCurrentEventRow (tablePosition.Position);
 		}
 
+		bool recalculatingHeight = false;
 		public void InvalidateRowHeight ()
 		{
-			source.InvalidateRowHeight (tablePosition.Position);
+			if (tablePosition != null && !recalculatingHeight) {
+				recalculatingHeight = true;
+				source.InvalidateRowHeight (tablePosition.Position);
+				recalculatingHeight = false;
+			}
+		}
+
+		public double GetRequiredHeightForWidth (double width)
+		{
+			Fill (false);
+			double height = 0;
+			foreach (var c in GetCells (new CGSize (width, -1)))
+				height = Math.Max (height, c.Frame.Height);
+			return height;
 		}
 
 		public override NSObject Copy ()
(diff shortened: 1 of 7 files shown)


Xwt.XamMac/Xwt.Mac.CellViews/CompositeCell.cs
```diff
@@ -88,9 +88,23 @@ public void SetCurrentEventRow ()
...
+				source.InvalidateRowHeight (tablePosition.Position);
+				recalculatingHeight = false;
+			}
+		}
+
+		public double GetRequiredHeightForWidth (double width)
+		{
+			Fill (false);
+			double height = 0;
+			foreach (var c in GetCells (new CGSize (width, -1)))
```
Therzok: Possible perf problem could arise because of `GetCells` being implemented via `IEnumerable` + `yield return`. We'll have one boxed allocation for this enumeration and another allocation for the iterator. This method is called in a loop.

Therzok: We know the size of the collection, so we can just return an Array of `CellPos` and then we would be allocationless.

sevoku: 👍 will change it

Xwt.XamMac/Xwt.Mac.CellViews/CompositeCell.cs
```diff
@@ -266,57 +263,48 @@ CGSize CalcSize ()
...
-		IEnumerable<CellPos> GetCells (CGSize cellSize)
+		List <CellPos> GetCells (CGSize cellSize)
 		{
 			int nexpands = 0;
 			double requiredSize = 0;
 			double availableSize = cellSize.Width;
 
-			var visibleCells = VisibleCells.ToArray ();
-			var sizes = new double [visibleCells.Length];
+			var cellFrames = new List<CellPos> (cells.Count);
```
Therzok: 👍

Xwt.XamMac/Xwt.Mac.CellViews/CompositeCell.cs
```diff
@@ -266,57 +263,48 @@ CGSize CalcSize ()
...
+					if (((ICellRenderer)cellFrames [i].Cell).Backend.Frontend.Expands)
+						cellFrames [i].Frame.Width += (nfloat)expandRemaining.NextSizePart ();
 				}
 			}
 
 			double x = 0;
-			for (int i = 0; i < visibleCells.Length; i++) {
-				var cell = (NSView)visibleCells [i];
-				var height = cell.FittingSize.Height;
+			for (int i = 0; i < cellFrames.Count; i++) {
```
Therzok: Style: can use foreach here now.

PR #801, opened 2018-03-09
[WPF] Move ExTreeViewItem brush styles from template to outer style

chkn:
This should not change the appearance or behavior, but should make it
easier to override these styles externally.


chkn:
Just found an issue.. will confirm and fix when I'm back at my Windows machine.

chkn:
Not sure if that actually made a difference, but I believe this is technically correct now.

PR #804, opened 2018-03-15
[Mac] Fix TreeView.ExpandToRow

sevoku:


Xwt.XamMac/Xwt.Mac/TreeViewBackend.cs
```diff
@@ -306,12 +307,19 @@ public void ScrollToRow (TreePosition pos)
 		public void ExpandToRow (TreePosition pos)
 		{
 			var p = source.GetParent (pos);
+			if (p == null)
+				return;
+			var s = new Stack<TreePosition> ();
```
Therzok: Why do we use a stack here?

sevoku: It was quick and easy. But we can replace it with a list if it's woth it perf wise.

Therzok: Nope, it's okay. Was wondering why we needed the `Stack<>` usage if it was just a linear list reverse.

PR #805, opened 2018-03-16
[A11y] Add ability to add native accessibility children to Xwt access…

iainx:
…ibles

Sometimes widgets need accessible children that aren't backed by individual
widgets, this allows them to be created natively and added to Xwt Widgets


iainx:
Realised I need to add ability to remove/clear the children too

PR #808, opened 2018-03-16
 Fallback to using the WPF MainWindow if can't get parent window

BretJohnson:
This fixes two scenarios for the Android designer:
- When the designer is hosted in a _floating_ VS window,
  GetParentWindow returns null.
 - When the designer is hosted in the new XET tabbed editor control
   GetParentWindow returns null (since it's actually a WinForms
   component,   not WPF, that contains the designer)

In both cases, falling back to the WPF Application.Current.MainWindow
makes things work propertly.


BretJohnson:
@garuma can you review

PR #813, opened 2018-03-20
[Mac] Cache cell width to improve size calculations

sevoku:
By caching the widths required to render each CompositeCell, the amount of required size calculations is reduced drastically. This also reduces the amount of drawing cycles and improves the overall speed and responsivness of lists/trees a lot.

**Limitation:** this adds caching only, but full cell resize is still not supported. A call to `QueueResize` will remeasure the cell and update the height required to render the cell with its current width. For a full resize we'll need to hook the column autosizing into the resizing process.

### DIFF
--- a/Xwt.XamMac/Xwt.Mac/ListViewBackend.cs
+++ b/Xwt.XamMac/Xwt.Mac/ListViewBackend.cs
@@ -213,7 +213,7 @@ namespace Xwt.Mac
 			};
 		}
 
-		public override void InvalidateRowHeight (object pos)
+		public override void QueueResizeRow (object pos)
 		{
 			UpdateRowHeight((int)pos);
 		}
@@ -223,7 +223,10 @@ namespace Xwt.Mac
 		{
 			if (updatingRowHeight)
 				return;
-			foreach (var colWidths in ColumnRowWidths)
+			
+			// FIXME: this won't resize the columns, which might be needed for custom cells
+			// In order to resize horizontally we'll need trigger column autosizing.
+			foreach (var colWidths in ColumnRowWidths) // invalidate widths for full recalculation
 				colWidths[(int)row] = -1;
 			RowHeights[(int)row] = CalcRowHeight (row);
 			Table.NoteHeightOfRowsWithIndexesChanged (NSIndexSet.FromIndex (row));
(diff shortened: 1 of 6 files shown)


Xwt.XamMac/Xwt.Mac/ListViewBackend.cs
```diff
@@ -140,14 +152,36 @@ void HandleDoubleClick (object sender, EventArgs e)
...
 			}
 		}
-		
+
+		public override NSTableColumn AddColumn (ListViewColumn col)
+		{
+			NSTableColumn tcol = base.AddColumn (col);
+			var widths = new List<nfloat> ();
+			if (Table.RowCount > 0)
+				widths.InsertRange (0, Enumerable.Repeat<nfloat> (-1f, (int)Table.RowCount));
```
Therzok: This will do Insert N times instead of doing a block insert. Avoid direct IEnumerables here. Please pass something `ICollection` derived. (the `IEnumerable<T>` overload handles checking for `ICollection<T>`)

Therzok: Actually, I think pre-allocating list capacity `(new List<nfloat> (table.RowCount)` then doing
`for (int i = 0; i < Table.RowCount; ++i) widths.Add (-1f)` is better

Xwt.XamMac/Xwt.Mac/ListViewBackend.cs
```diff
@@ -140,14 +152,36 @@ void HandleDoubleClick (object sender, EventArgs e)
...
 		public virtual void SetSource (IListDataSource source, IBackend sourceBackend)
 		{
 			this.source = source;
 
 			RowHeights = new List<nfloat> ();
 			for (int i = 0; i < source.RowCount; i++)
 				RowHeights.Add (-1);
+			foreach (var colWidths in ColumnRowWidths) {
+				colWidths.Clear ();
+				colWidths.AddRange (Enumerable.Repeat<nfloat> (-1, source.RowCount));
```
Therzok: Same as above, AddRange on `IEnumerable<T>` but not `ICollection<T>` does not do block copy.

PR #816, opened 2018-03-26
[WPF] Fix ReadOnly property on TextEntryBackend

### LINKED ISSUE #815: [Wpf] TextEntry.ReadOnly error in wpf backend
If once set Text Entry.ReadOnly property it become always true even if set to false

in Xwt.WPFBackend.TextEntryBackend
now:
```
        public bool ReadOnly
	{
		get { return TextBox.IsReadOnly; }
		set { TextBox.IsReadOnly = true; }
	}
```
but expected to be:
```
        public bool ReadOnly
	{
		get { return TextBox.IsReadOnly; }
		set { TextBox.IsReadOnly = value; }
	} 
```

Therzok:
Issue pointed out by @alexandrvslv in #815

Fixes #815


Therzok:
Audited other bits of code, seems like nothing else uses it.

PR #818, opened 2018-03-27
[Mac] Fix cast exception

paymicro:
When an image is drawn, there may be a cast exception because backend may not be an instance of the `NSImage` class.

https://prnt.sc/iwn2oq


sevoku:
This is the expected behaviour, and the proposed check would hide the real issue (image being drawn in a wrong context). When using images in a cross-toolkit environement, Xwt will verify the image backend and convert to the current backend if necessary (when using Xwt APIs). To manually render an image with a specific toolkit directly, you'll need to get the native image using `Toolkit.GetSafeBackend`.
@paymicro which particual issue are you trying to solve?

paymicro:
@sevoku one extension for Xamarin iOS designer throws this exception. There is a suggestion that it causes a long boot designer. Since it is an expected behavior then may be it is only a consequence.

PR #822, opened 2018-04-10
D15 7 registrar fixes

Therzok:


sevoku:
#821 contains also e26e3c540c4b2cca42b054c83b42f0aae9bc5794, what about this PR?

Therzok:
Fixed.

alanmcgovern:
💯

Xwt.XamMac/Xwt.Mac/TextEntryBackend.cs
```diff
@@ -356,6 +356,29 @@ public CustomCell ()
...
+			/// which makes it easier to debug issues.
+			/// </summary>
+			/// <returns>The copy.</returns>
+			/// <param name="zone">Zone.</param>
+			public override NSObject Copy(NSZone zone)
+			{
+				// Don't remove this override because the comment on this explains why we need this!
+				var newCell = (CustomCell)base.Copy(zone);
+				newCell.editor = editor;
+				newCell.selChangeObserver = selChangeObserver;
```
alanmcgovern: Do we want to copy an event registration from *a different c# object* to this C# object? When the event is raised the `HandleSelectionDidChange` method will be invoked on the *original* object only, not on the clone.

alanmcgovern: (I ask because it'll be confusing/weird to have a non-null field. it won't really cause a bug though).

Therzok: Doesn't copying also copy everything else?

alanmcgovern:
It’s the same as a memberwise clone in .NET. When you do one of those in c# you may have to null out fields if copying the old object makes no sense. In this case copying an event registration token for an old object makes no sense, neither does creating a new event registration for the new clone :)

Therzok:
I'll open PRs to revert that change.

alanmcgovern:
It's not super important for this specific case, but figuring out the correct subset of managed state to copy is important to keep in mind in general :)

PR #826, opened 2018-04-27
[Wpf] Don't skip the Measure step all the time

alanmcgovern:
If we add something to the live tree it will have
an invalid measure and arrange. The first measure pass
will run as-expected.

If you remove and re-add the widget to the live tree
then it's measure and arrange will be invalidated, but
because we do not call 'ArrangeChildren (force = true)'
then we will never invoke the Measure for the children
again.

For the case of ListBox this means an implciit measure
will occur, which passes Inf,Inf as the size, which
means we lose scrollbars as it expands to fill all
that space.

This fixes the ExceptionPopover in the designers.

### DIFF
--- a/Xwt.WPF/Xwt.WPFBackend/BoxBackend.cs
+++ b/Xwt.WPF/Xwt.WPFBackend/BoxBackend.cs
@@ -129,9 +129,8 @@ namespace Xwt.WPFBackend
 					if (force) {
 						// Don't recalculate the size unless a relayout is being forced
 						element.InvalidateMeasure ();
-						element.Measure (new SW.Size (r.Width, r.Height));
 					}
-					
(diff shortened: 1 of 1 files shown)


alanmcgovern:
For reference this is the construct that breaks: https://github.com/xamarin/designer/blob/a3187b38f0bfe75ef2a93a3932eb5f8c4c56cf4a/Xamarin.Designer/Xamarin.Designer.VisualItems/ExceptionCallout.cs#L71-L120

And it breaks because both the measure and arrange are marked as invalid by WPF but the original code only ever called Arrange.

alanmcgovern:
Screencast of the fix, then i undo the fix: http://recordit.co/W4AjQ3Qrsl

PR #828, opened 2018-05-03
[Mac] Fixed `NRE` converting points from events.

viniciusjarina:
a NSView without a Window will cause `ConvertPointFromEvent` to throw a `NRE`, in this case is better just skip the BaseToScreen conversion.


alanmcgovern:
What triggers this? Do you have the stacktraces from the two cases? I'm wondering how we can be getting mouse events for something that is *not* in a window. While we should probably fix this anyway, i wonder if we also need to patch something else in our codebase to not hit this case in the first place :)

alanmcgovern:
pingo :) 

We might need this in master and 15.7 and 15.8!

PR #831, opened 2018-05-16
[WPF] Accessibility support

garuma:
This PR expands the accessibility support for the WPF backend in those ways:

- Add `LabeledBy` API to allow another widget to label this one
- Support the children accessibility API (for `Canvas` only as WPF doesn't allow to modify this information for existing widgets in general without subclassing and we mostly only care about supplying children accessibility object for custom-drawn and thus Canvas based elements anyway).
- Add a few more automation peers to some custom WPF controls we have

### DIFF
--- a/Xwt.WPF/Xwt.WPFBackend/AccessibleBackend.cs
+++ b/Xwt.WPF/Xwt.WPFBackend/AccessibleBackend.cs
@@ -51,8 +51,7 @@ namespace Xwt.WPFBackend
 
 		public void Initialize (IWidgetBackend parentWidget, IAccessibleEventSink eventSink)
 		{
-			var parent = parentWidget as WidgetBackend;
-			Initialize (parent.Widget, eventSink);
+			Initialize (parentWidget.NativeWidget, eventSink);
 		}
 
 		public void Initialize (object parentWidget, IAccessibleEventSink eventSink)
--- a/Xwt.WPF/Xwt.WPFBackend/AccessibleBackend.cs
+++ b/Xwt.WPF/Xwt.WPFBackend/AccessibleBackend.cs
@@ -31,7 +31,7 @@ namespace Xwt.WPFBackend
 
 		public string Title { get; set; }
 		public string Value { get; set; }
-		public Role Role { get; set; }
+		public Role Role { get; set; } = Role.Custom;
 		public Uri Uri { get; set; }
 		public Rectangle Bounds { get; set; }
 		public string RoleDescription { get; set; }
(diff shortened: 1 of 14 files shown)


Xwt.WPF/Xwt.WPFBackend/AccessibleBackend.cs
```diff
@@ -46,31 +53,87 @@ public void EnableEvent (object eventId)
...
+		{
+			switch (role) {
+			case Role.Button:
+			case Role.MenuButton:
+			case Role.ToggleButton:
+				return AutomationControlType.Button;
+			case Role.CheckBox:
+				return AutomationControlType.CheckBox;
+			case Role.RadioButton:
+			case Role.RadioGroup:
```
sevoku: Shouldn't `Role.RadioGroup` map to `AutomationControlType.Group`?

Xwt.WPF/Xwt.WPFBackend/AccessibleBackend.cs
```diff
@@ -46,31 +53,87 @@ public void EnableEvent (object eventId)
...
+			case Role.Label:
+				return AutomationControlType.Text;
+			case Role.Link:
+				return AutomationControlType.Hyperlink;
+			case Role.Image:
+				return AutomationControlType.Image;
+			case Role.Cell:
+				return AutomationControlType.DataItem;
+			case Role.Table:
+				return AutomationControlType.Table;
```
sevoku: `AutomationControlType.Table` indicates that the table contains only text, I think this should be [`AutomationControlType.DataGrid`](https://docs.microsoft.com/de-de/dotnet/api/system.windows.automation.controltype.datagrid?view=netframework-4.7.1#remarks) because we support any cell type.

Xwt/Xwt.Accessibility/Accessible.cs
```diff
@@ -120,6 +120,12 @@ internal Accessible (object nativeParent)
 			}
 		}
 
+		public Widget LabeledBy {
```
sevoku: This naming is a little bit confusing, I think something like `LabelWidget` would be better. @iainx afair there was something similar in Cocoa, do you remember how it works?

sevoku: Also why doesn't it have a getter?

iainx: There are two properties in Cocoa

`AccessibilityTitleUIElement` which links the control element to its title control
`AccessibilityServesAsTitleForUIElements` which links the title control to an array of control elements

So `LabeledBy` would be the equivalent of `AccessibilityTitleUIElement`

garuma: Cool, I'll change the name and add the Cocoa side.

No getter as there is no way to map the Xwt.Widget back in a practical way from the underlying implementation. I can however store the value in `Accessible` directly if that's useful?

Xwt/Xwt/Widget.cs
```diff
@@ -332,6 +332,10 @@ protected override void Dispose (bool disposing)
 				return accessible;
 			}
 		}
+
+		public bool HasAccessible {
```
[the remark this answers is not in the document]
garuma: Sure, I can do it like that instead

Xwt/Xwt.Accessibility/Accessible.cs
sevoku: @garuma yeah, but in this case there is no way from WPF to Xwt (or do some WPF widgets have a predefined `LabeldBy` UIElement?), so to be aligned with the rest of the API there should be a getter and we could store a reference in `Accessible`. This would make it possible to check whether there is one already assigned and which one.

garuma:
@sevoku I think I addressed all the existing feedback, anything else to change?

PR #839, opened 2018-06-25
More WPF accessibility changes

BretJohnson:
This PR has more changes to support accessibility for the Android Designer and our other Xwt WPF clients.

See the individual commit messages for a description of the five changes.


Xwt/Xwt/Popover.cs
```diff
@@ -84,7 +85,17 @@ public Popover (Widget content)
 			VerifyConstructorCall (this);
 			Content = content;
 		}
-		
+
+		Accessible accessible;
```
garuma: fields at the top of the file

Xwt/Xwt/ColorPicker.cs
```diff
@@ -120,7 +120,7 @@ protected virtual void OnColorChanged (EventArgs args)
 		}
 	}
 
-	class DefaultColorPickerBackend: XwtWidgetBackend, IColorPickerBackend
+	public class DefaultColorPickerBackend: XwtWidgetBackend, IColorPickerBackend
```
garuma: If there are a11y naming to be done it makes more sense to do them here directly instead of letting the client do it.

Xwt.WPF/Xwt.WPFBackend/WindowsSpinButton.xaml.cs
```diff
@@ -458,9 +459,19 @@ private void textBox_MouseWheel(object sender, MouseWheelEventArgs e)
...
 		{
-			return new WindowsSpinButtonAutomationPeer (this);
+			WindowsSpinButton spinButton;
+
+			public SpinButtonTextBox (WindowsSpinButton spinButton)
+			{
+				this.spinButton = spinButton;
+			}
+
+			protected override AutomationPeer OnCreateAutomationPeer ()
```
garuma: This is not correct, the automation peer needs to be on the overall user control not the inner textbox. It should then not advertise any children element. This is consistent with the rest of the platform and what I was told by the accessibility people.

Xwt/Xwt/Popover.cs
BretJohnson: I just copied here what's in Widget for `Accessible`. But let me update both of them.

Xwt/Xwt/ColorPicker.cs
BretJohnson: Here's how I use this: https://github.com/xamarin/designer/blob/207acc07d5e54edd3a4419247b7a740d581300b0/Xamarin.Designer.Android/Xamarin.AndroidDesigner/Xamarin.AndroidDesigner.Selectors/GridOverlaySelector.cs#L255

Note that the a11y name here for the button that launches the color picker comes from the tooltip text for us, which needs to come from the client as it's client specific. There might be a more ideal way to handle this, supporting setting the Accessible name on the frontend ColorPicker instead of the special backend-which-lives-in-the-frontend DefaultColorPickerBackend, but it seemed a bit tricky to plumb thru the Accessible in that special case & this worked fine.

Separately, though, see my subsequent changes in this PR for `ColorSelector.cs`, where I do add a11y names to Xwt for the color related UI (e.g, the RGB and HSL editors).

Xwt.WPF/Xwt.WPFBackend/WindowsSpinButton.xaml.cs
BretJohnson: It just didn't work before, when the parent UserControl returned the peer, as it's the TextBox that has the focus and is thus what's queried by e.g. the Narrator for the peer. The spinner control is "mostly" a textbox, so there's some logic in doing this, though we may want to change the parent control to be "pane" instead of "custom".

Anyway, let me ask at office hours today about this. There might be a better way to handle it.

BretJohnson: I ended up changing changed this slightly. Now the outer spinner UserControl and the TextBox both return the same automation peer, which is associated with the outer spinner. Again, with the original code the Narrator skipped over the spinner controls, not apparently seeing them as accessible when the TextBox inside had the focus. But this changed version seems to work well.

And FYI - I did attend office hours. Guy Barker is the person who wrote much of the Windows automation API. He wasn't there when I attended, but I got his name. I didn't end up reaching out to him for this, but we may reach out in the future.

BretJohnson:
With the last commit, this should be good to merge now.

PR #840, opened 2018-07-06
[WPF] Only take border size into account when decorated

imerr:
I was trying to make an unstyled window movable via dragging the background and the window would move even when setting the location to itself, after some digging and trial & error I discovered GetBorderSize() would still return a size, even though there was none

This would cause the window to move by 3px in both x/y when setting the Location/calling Move

I'm not 100% sure this is the right way to fix it (you might want to change GetBorderSize?), but it certainly works and fixes the issue. I'll gladly change things around if anyone wants to suggest a better way

I've only tested it on Windows 10


sevoku:
Thanks!

PR #846, opened 2018-08-21
[Mac][Icons] New zoom icons

vancura:
Here's new set of zoom icons:

zoom

As you can see they're much simpler now (I removed the lens symbol). For that reason, we still need to test if the icons are still readable enough sporting this minimalism. Hence, please don't merge yet.


vancura:
This is now approved by the design team. Can you please review please @slluis? PRs in our other repos will come as soon as this is in master.

PR #862, opened 2018-10-10
[Gtk] Make CellRenderers activatable

Therzok:
This change allows treeview/listview toggle cells to not be toggled,
unless the toggle button is clicked.


Therzok:
Sevo verified it also. Merging.

PR #863, opened 2018-10-16
[WPF] Indicate ScrollView keyboard focus

daserge:


daserge:
scroll_view_focus

daserge:
Android SDK Manager indeed doesn't react to VS' theme, only to system-wide High-Contrast mode, which looks like this:
windows_high_contrast

PR #864, opened 2018-10-18
[a11y] Add accessible API for Menu and MenuItem

DmitriyKirakosyan:


sevoku:
The `Menu` code lgtm, but what about the `MenuItem`?

Is the `TreeStoreNode` change in 5bb63a4ad3461a1386ecdfccf458cdc7777d40f0 related to menus? Which problem does it solve?

Also please avoid formatting changes, since that make it difficult to see relevant changes in the diff.

sevoku:
The `Menu` code lgtm, but what about the `MenuItem`?

Is the `TreeStoreNode` change in 5bb63a4ad3461a1386ecdfccf458cdc7777d40f0 related to menus? Which problem does it solve?

Also please avoid formatting changes, since that make it difficult to see relevant changes in the diff.

sevoku:
The `Menu` code lgtm, but what about the `MenuItem`?

Is the `TreeStoreNode` change in 5bb63a4ad3461a1386ecdfccf458cdc7777d40f0 related to menus? Which problem does it solve?

Also please avoid formatting changes, since that make it difficult to see relevant changes in the diff.

sevoku:
The `Menu` code lgtm, but what about the `MenuItem`?

Is the `TreeStoreNode` change in 5bb63a4ad3461a1386ecdfccf458cdc7777d40f0 related to menus? Which problem does it solve?

Also please avoid formatting changes, since that make it difficult to see relevant changes in the diff.

sevoku:
The `Menu` code lgtm, but what about the `MenuItem`?

Is the `TreeStoreNode` change in 5bb63a4ad3461a1386ecdfccf458cdc7777d40f0 related to menus? Which problem does it solve?

Also please avoid formatting changes, since that make it difficult to see relevant changes in the diff.

daserge:
https://github.com/mono/xwt/commit/5bb63a4ad3461a1386ecdfccf458cdc7777d40f0 makes Windows Narrator to announce tree row label via overriden `TreeStoreNode.ToString`. @DmitriyKirakosyan please correct me if I'm wrong.

As for whitespaces adding `?w=1` really helps in this case:
https://github.com/mono/xwt/commit/5bb63a4ad3461a1386ecdfccf458cdc7777d40f0?w=1
Although these changes kind of spoil the history so makes sense to avoid it if possible.

sevoku:
In https://github.com/mono/xwt/commit/5bb63a4ad3461a1386ecdfccf458cdc7777d40f0?w=1 Is only see the `Name` property randomly added to some classes incl. public API. I'm not sure how it will solve the actual problem, however it should have a clear API and go to a separate PR, since it's unrelated to the Menus.

DmitriyKirakosyan:
@sevoku Basically the issue was in WPF only (narrator announced list items as 'tree view, Xwt.WPFBackend.TreeStoreNode). So it fixes the problem because `TreeStoreNode.ToString` is used by `Name` accessible property. So Narrator reads it correctly with these changes.

However, I'm going to revert this commit, since I'm still working on it and going to add more code related to the tree view a11y, which would be better to send as a different PR. Also currently it does look weird, because the `Name` property is really needed for Windows only. In my recent changes it is already reworked.

Xwt.XamMac/Xwt.Mac/AccessibleBackend.cs
```diff
@@ -54,6 +54,11 @@ public void Initialize (IPopoverBackend parentPopover, IAccessibleEventSink even
 			// Not currently supported
 		}
 
+		public void Initialize (IMenuBackend parentMenu, IAccessibleEventSink eventSink)
+		{
+			Initialize (parentMenu, eventSink);
```
sevoku: cast to `NSMenu` to call the other overload, otherwise it's infitie recursion

garuma:
Assuming things are good now that the last change was added. Feel free to revert the merge if not.

DmitriyKirakosyan:
🎉

PR #867, opened 2018-10-25
[a11y] Add accessible API to TreeView items.

DmitriyKirakosyan:
* TreePosition.Accessible
* TreePosition.Accessible.AddChild can be used for setting up custom automation peers for TreeView and TreeViewItem


DmitriyKirakosyan:
@BretJohnson @sevoku  Please review https://github.com/mono/xwt/pull/867/commits/c1b79046ebf7fee20ac5c21ccdd9cf160266b6d7.

I had to replace the `AutomationPeer` logic in `ExTreeViewItem` with the IAutomationPeerCreator because a client can't create a peer, as it doesn't know about `ExTreeViewItem` native item and can't get it from `TreeView` and `TreeNavigator` APIs.
Please let me know if I missed something or did something wrong.

I will push a PR to sdk-installer soon that will demonstrate the usage of this PR.

Xwt.WPF/Xwt.WPFBackend/ExListView.cs
```diff
@@ -75,5 +77,18 @@ internal void FocusItem(ListViewItem item)
...
+
+		protected override AutomationPeer OnCreateAutomationPeer ()
+		{
+			var backend = this.Backend as ListViewBackend;
+			var listView = backend?.Frontend as Xwt.ListView;
+
+			if (listView != null) {
+				var peers = listView.Accessible.GetChildren ();
+				if (peers.Any () && peers.First () is AutomationPeer)
+					return peers.First () as AutomationPeer;
```
daserge: You can use direct cast here instead of `as` (the check is already done above)

Xwt.Gtk/Xwt.GtkBackend/AccessibleBackend.cs
```diff
@@ -192,6 +193,12 @@ public virtual void RemoveAllChildren ()
 			// TODO
 		}
 
+		public virtual IEnumerable<object> GetChildren ()
+		{
+			// TODO
+			return null;
```
slluis: This should return an empty enumeration by default, not null.

Xwt/Xwt/TreeNavigator.cs
```diff
@@ -115,6 +115,17 @@ public TreeNavigator AddChild ()
 			pos = backend.AddChild (pos);
 			return this;
 		}
+
+		public TreeNavigator GetChild (int index)
```
slluis: This should be `MoveToChild(int index)`

Xwt/Xwt/TreePosition.cs
```diff
@@ -26,13 +26,22 @@
...
 using System.Collections.Generic;
+using Xwt.Accessibility;
 
 namespace Xwt
 {
-	public interface TreePosition
+	public class TreePosition
 	{
+		Accessible accessible;
+		public Accessible Accessible {
```
slluis: This is not right. TreePosition is just a reference to a position, it can't contain data itself.

DmitriyKirakosyan: @slluis,
Hmm. I think alternatively I can add a11y API to `ITreeStoreBackend`, which would look like this:
```
ITreeStoreBackend {
    Accessible GetChildAccessibility (TreePosition child);
} 
```
Does it look good to you?

DmitriyKirakosyan: @slluis I realized it might be not a good idea to instantiate Accessibility on the backend side, moreover its constructors are internal.

The problem is I need to send AutomationPeer somehow to the WPF backend, using `TreePosition`, `TreeNavigator` or `TreeStore` APIs. If I don't use TreePosition, then I should use `TreeStore` or`TreeNavigator` to send it to ITreeStoreBackend, where I would attach the accessible object to the proper Child (TreeStoreNode in case of Xwt.WPF). The API would like like this `TreeNavigator.AddAccessibleChild (object nativeChild)` (and maybe `RemoveAccessibleChild`), which would be used to deliver `AutomationPeer` to `TreeStoreNode` and then to `ExTreeViewItem`.

Or also I can use `TreePosition` interface for the same purpose -- `TreePosition.AddAccessibleChild (object nativeChild)`. This way the accessible object would be delivered directly to `TreeStoreNode`, without redirecting this work to `TreeStoreBackend`.
What do you think?

sevoku: I'm not convinced that this is the right approach in general. `TreePosition`, `TreeNavigator` or `TreeStore` are handling the data and not it's representation, hence they can not be accessible in any way, `TreeView` with it's `CellView`s is the visualization and can/should have accessibility features.

The right approach might be to add `Accessible` and `AccessibleField` properties to the `CellView`s. It could be either a `DataField<Accessible>` or we could map several `Accessible*` fields (AccessibleLabelField, etc.). this would allow us to control accessibility on cell level rather than on rows (your WPF specific approach is row based, since `ExTreeViewItem` is a single row representation). This is also how Cocoa accessibility works (setting the props on single NSViews in view based tree mode).

slluis: I agree with @sevoku. I think a good approach is to have an AccessibleFields class in CellView, which would have one field for each accessible property. Something like this:

``` c#
public class CellView {
    ...
    public AccessibleFields AccessibleFields { get; } = new AccessibleFields ();
    ...
}

public class AccessibleFields
{
    public DataField<string> Label = new DataField<string> ();
    public DataField<string> Title = new DataField<string> ();
    ....
}
```

It would be better to crate all this on demand though.

DmitriyKirakosyan: @sevoku Hm, but TreeView doesn't expose any `CellView`s. And even if we can get `CellView` somehow, settings a11y to specific CellView inside the row wouldn't work for the whole row. We are selecting the whole row and see its a11y properties and not its internal UI components. So user would need to go inside the row through its components to get a picture what is presented in that row.  I might get something wrong here, correct me if so.
Getting `CellView` might be related to the `FrameworkElementFactory`, which is used to create the column cell template: https://github.com/mono/xwt/blob/master/Xwt.WPF/Xwt.WPFBackend/TreeViewBackend.cs#L261. Not sure how exactly, but maybe by listening events from this factory. And then we should expose these `CellView`s to the Xwt client using something like Xwt.CellView as a fronted representation. Is it what you meant?
If so, don't we need to set `Accessibility` to ExTreeViewItem as well? That would mean that we need to expose TreeViewItem's frontend like Xwt.TreeViewItem?

sevoku: The thing is that a row can contain several completely independent views. Not only column-wise, also every column can contain several CellCiews inside. Pls check the samlples especially here: https://github.com/mono/xwt/blob/master/TestApps/Samples/Samples/TreeViews.cs#L48-L102.
So if a user selects a row and we present "some" accessible text for it, it can't reflect all the information inside the particular row, since it can contain various different components, represented ba different `CellView` types (besides a label, it can be a checkbox or a combo box or whatever).
In Cocoa and also WPF a row contains different views, arranged either by columns or specialized views inside a column (a column can contain several different view types). Each view (text, icon, check, combo,...) has different purposes and information. Hence it is not sufficient to present just a row-based information, which does not allow the user to interact with different components and also get column specific information.
@DmitriyKirakosyan you are talking about WPF specific "hacks", to make single rows accessible, which is not sufficient to represent all data in a view context and is not compatible with other toolkits (only Xwt.WPF is using a special row-view) . And what you propose is data-driven and not view related.

DmitriyKirakosyan: @sevoku Thanks for the explanation! It does make sense to me now. Just to make sure I understand you correctly:
I add `AccessibleFields` to CellView, then on the client side I add a11y fields to a `CellView` object and to `TreeStore` here https://github.com/mono/xwt/blob/master/Xwt/Xwt/TreeStore.cs#L59, and then set values here https://github.com/mono/xwt/blob/master/Xwt/Xwt/TreeNavigator.cs#L119
Right?

sevoku: Yeah, this is exactly what I mean :) So it works like all the other CellView data/properties.

DmitriyKirakosyan:
@sevoku Please review ^

`CellView` Accessibility seems working for `Xwt.Mac` and `Xwt.WPF` fine with these changes.
Notice this thing: https://github.com/mono/xwt/pull/867/commits/a70641bf986c4cd94ef392b5e701b6ec39f8e1cf#diff-469f2a40e1b6cdfeaaf5cdbd086eb033R266.
I'm not sure if I do it correctly, but I don't see any other good way to handle this. We need the row children to be accessible and also we don't want to mess them with `TreeViewItem` children. So It might be reasonable to hide the row UI items when it is expanded in order to correctly display the shown children number.

I didn't find a way to support `Xwt.Gtk`, though. The solution should be around `CustomCellRenderer.cs`, as far as I understand. I will investigate more.

DmitriyKirakosyan:
@sevoku I extracted the CellView changes to the new PR https://github.com/mono/xwt/pull/897. Please review when you have a moment.

I'm wondering if any of other changes from this PR are needed. F.e. https://github.com/mono/xwt/pull/867/files#diff-82439351aac27f9eb2a87bba5d8b2443R150, this was done for allowing a client to set custom AutomationPeer to WPF native objects. Using here: https://github.com/mono/xwt/pull/867/files#diff-3d9a78c59e4d79eb6619e47a53e3b0a7R484 and here https://github.com/mono/xwt/pull/867/files#diff-9aa67e58bff9459fe290865599b0e4efR89. 

Also these changes https://github.com/mono/xwt/pull/867/files#diff-798533a5d4653f71651d0fde0885384bR112, which are for those native widgets which don't exist at the moment of setting its a11y properties. While native widget is not created all a11y properties are set to default backend, and then (when native widget is accessed) migrate to a real one: https://github.com/mono/xwt/pull/867/files#diff-798533a5d4653f71651d0fde0885384bR139.

I personally don't need these changes, but maybe some of them could be useful, especially the changes for AutomationPeer customization.

PR #875, opened 2018-11-15
[Gtk][a11y] Support a11y for combobox entry

DmitriyKirakosyan:
Setting Accessibility to `Xwt.ComboBoxEntry` doesn't work for `Gtk`, because `ComboBoxEntryBackend` uses `Gtk.Bin` as a container for entry ui element (https://github.com/mono/xwt/blob/master/Xwt.Gtk/Xwt.GtkBackend/ComboBoxEntryBackend.cs#L37).
So `Widget.Child` in this case is Entry element, which should be used as an a11y target.

### DIFF
--- a/Xwt.Gtk/Xwt.GtkBackend/AccessibleBackend.cs
+++ b/Xwt.Gtk/Xwt.GtkBackend/AccessibleBackend.cs
@@ -24,6 +24,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 using System;
+using System.Linq;
 using Xwt.Accessibility;
 using Xwt.Backends;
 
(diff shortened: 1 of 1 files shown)


sevoku:
LGTM, but I wonder why Gtk/Atk don't support that our of the box, I'd assume that setting `Accessible` props on Gtk.ComboBox would do the right thing. How does it work on Linux?

iainx:
Yeah, AtkCocoa doesn't handle Combobox Entries at the moment. I need to look and see how a similar control is handled in Cocoa

DmitriyKirakosyan:
@sevoku please check ^
Also, there were some doubts about whether we should choose `Gtk.Entry` or `Gtk.ToggleButton` as an a11y target for the `ComboBoxEntry`.
Just discussed this with QA and got approval that it is fine to set a11y props to `Gtk.Entry` and ignore the toggle button.

Xwt.Gtk/Xwt.GtkBackend/AccessibleBackend.cs
```diff
@@ -24,6 +24,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 using System;
+using System.Linq;
```
sevoku: Is this using required?

PR #893, opened 2018-11-27
Adds Size Attribute parsing in Span tags

netonjm:
This PR adds implementations  to cover "size" attribute in span tags in Gtk and Cocoa backends,

Example:

`<span size=\"10720\" font=\"Tahoma\" color='#0000ff'>blue</span>`

`<span size=\"xx-small\" font=\"Arial\" color='#ff0000'>red</span>`

`<span size=\"xx-large\" font=\"Tahoma\" color='#00ff00'>green</span>`

image

image-2

### DIFF
--- a/Xwt/Xwt.Drawing/FontSizeTextAttribute.cs
+++ b/Xwt/Xwt.Drawing/FontSizeTextAttribute.cs
@@ -32,8 +32,15 @@ namespace Xwt.Drawing
 	public sealed class FontSizeTextAttribute : TextAttribute
 	{
 		const float PangoScale = 1024;
-		public const string XXSmall = "xx-small", XSmall = "x-small", Small = "small", Medium = "medium", Large = "large", XLarge = "x-large", XXLarge = "xx-large",
-			Smaller = "smaller", Larger = "larger";
+		public const string XXSmall = "xx-small",
+			XSmall = "x-small", 
+			Small = "small", 
+			Medium = "medium", 
+			Large = "large", 
+			XLarge = "x-large", 
+			XXLarge = "xx-large",
+			Smaller = "smaller", 
+			Larger = "larger";
 
 		public float Size { get; private set; }
 		public string OriginalSizeStringValue { get; private set; }
--- a/Xwt/Xwt.Drawing/FontSizeTextAttribute.cs
+++ b/Xwt/Xwt.Drawing/FontSizeTextAttribute.cs
@@ -1,0 +1,87 @@
+﻿//
+// FontWeightTextAttribute.cs
+//
+// Author:
+//       Lluis Sanchez <lluis@xamarin.com>
+//
+// Copyright (c) 2013 Xamarin Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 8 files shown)


Xwt/Xwt.Drawing/FontSizeTextAttribute.cs
```diff
@@ -0,0 +1,87 @@
...
+		}
+
+		public void SetSize (string value)
+		{
+			OriginalSizeStringValue = value;
+
+			var currentSize = Font.SystemFont.Size;
+
+			float size;
+			if (float.TryParse (value, out size)) {
```
sevoku: Hhm, I think that we should support both Pango values (pt * 1024) and absolute pt. TBH Pango values are really uncommon, so just for compatibility with pango markup. We could try parse as float and then check if it's `<PangoScale`. For smaller values, just use the size, or divide otherwise.

Xwt/Xwt.Drawing/FontSizeTextAttribute.cs
```diff
@@ -0,0 +1,87 @@
...
+// THE SOFTWARE.
+
+using System.Linq;
+using System.Collections.Generic;
+
+namespace Xwt.Drawing
+{
+	public sealed class FontSizeTextAttribute : TextAttribute
+	{
+		public const float MaxSize = 1024;
```
sevoku: Let's call it `PangoScale` and make it private. This should only be for compatibility with Pango and the regular case should be PT.

Xwt/Xwt.Drawing/FontSizeTextAttribute.cs
```diff
@@ -0,0 +1,87 @@
...
+			OriginalSizeStringValue = value;
+
+			var currentSize = Font.SystemFont.Size;
+
+			float size;
+			if (float.TryParse (value, out size)) {
+				Size = size / MaxSize; //we need convert this to pt. values
+				return;
+			}
+			if (SizeAbsoluteValues.TryGetValue (value,out size)) {
```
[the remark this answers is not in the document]
sevoku: @slluis if we assume that those strings are Pango compatible, then they actually are absolute according to https://developer.gnome.org/pango/stable/PangoMarkupFormat.html

Xwt/Xwt.Drawing/FontSizeTextAttribute.cs
```diff
@@ -0,0 +1,88 @@
...
+
+using System.Linq;
+using System.Collections.Generic;
+
+namespace Xwt.Drawing
+{
+	public sealed class FontSizeTextAttribute : TextAttribute
+	{
+		const float PangoScale = 1024;
+		public const string XXSmall = "xx-small", XSmall = "x-small", Small = "small", Medium = "medium", Large = "large", XLarge = "x-large", XXLarge = "xx-large",
```
Therzok: Nit: formatting here could be better, one per line.

Therzok:
Looks good to me

Xwt/Xwt.Drawing/FontSizeTextAttribute.cs
netonjm: done

PR #897, opened 2018-12-02
[a11y][WPF][Mac] Add Accessibility API to CellView

DmitriyKirakosyan:
Moved from https://github.com/mono/xwt/pull/867.

Xwt.Gtk is not supported. Gtk.CellRenderer doesn't provide a11y API.

### DIFF
--- a/Xwt/Xwt/CellView.cs
+++ b/Xwt/Xwt/CellView.cs
@@ -27,6 +27,7 @@
 using System;
 using Xwt.Drawing;
 using Xwt.Backends;
+using Xwt.Accessibility;
 using System.ComponentModel;
 using System.Collections.Generic;
 
@@ -164,6 +165,8 @@ namespace Xwt
 			get { return GetValue (VisibleField, visible); }
 			set { visible = value; }
 		}
+
+		public AccessibleFields AccessibleFields { get; set; }
 
 		ICellViewEventSink ICellViewFrontend.Load (ICellDataSource dataSource)
 		{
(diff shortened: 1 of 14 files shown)


Xwt/Xwt/CellView.cs
```diff
@@ -165,6 +166,8 @@ public void OnButtonReleased (ButtonEventArgs args)
 			set { visible = value; }
 		}
 
+		public AccessibleFields AccessibleFields { get; set; }
```
slluis: Please change to:

```c#
AccessibleFields accessibleFields;

public AccessibleFields AccessibleFields {
    get {
        if (accessibleFields == null)
            accessibleFields = new AccessibleFields ();
        return accessibleFields;
    }
}
```

DmitriyKirakosyan: @slluis I would also need to remove these checks:
https://github.com/mono/xwt/pull/897/files#diff-caf37f9c62d1d45f8d30ad776d8b62d7R196
https://github.com/mono/xwt/pull/897/files#diff-2229d0433bd649eface587819a9c0a4cR190

Which would result in executing all the code in these methods even if the `AccessibilityFields` wasn't initialized.
Do you think it is still worth doing this change?

slluis: Yes, that change is necessary. ICellViewFrontend.AccessibleFields could still return null if the accessibleFields field has no been initialized.

DmitriyKirakosyan: 👌

PR #898, opened 2018-12-03
Toolkit guest

iainx:
Allow Xwt to be initialised as a guest inside Cocoa

### DIFF
--- a/Xwt/Xwt/Toolkit.cs
+++ b/Xwt/Xwt/Toolkit.cs
@@ -190,7 +190,7 @@ namespace Xwt
 		/// <param name="fullTypeName">The <see cref="Type.FullName"/> of the toolkit type.</param>
 		public static Toolkit Load (string fullTypeName)
 		{
-			return Load (fullTypeName, true);
+			return Load (fullTypeName, true, true);
 		}
 
 		/// <summary>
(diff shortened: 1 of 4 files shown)


Therzok (approved):

sevoku (approved):
Looks good to me!
@slluis could you take a look? You disabled it in https://github.com/mono/xwt/commit/1cb1e13f0aea6d723df13ac23bbbffebaefac0e3, do you remember why?

slluis:
I don't remember the reason. What is this needed for?

iainx:
Nothing in MonoDevelop. I just wanted to use Xwt in a Cocoa application and I couldn't because the guest feature had been disabled so it was trying to initialise NSApplication twice

slluis:
OK o me, as long as it doesn't break MD.

iainx:
@slluis @sevoku the newest commit means MonoDevelop won't break, I missed one overload :)

iainx:
Updated patch for new master

Xwt/Xwt/Toolkit.cs
```diff
@@ -224,15 +224,26 @@ internal static Toolkit Load (string fullTypeName, bool isGuest)
 		/// Load a toolkit of a specified type.
 		/// </summary>
 		/// <param name="type">The toolkit type.</param>
-		public static Toolkit Load (ToolkitType type)
+		public static Toolkit Load(ToolkitType type)
+		{
+			return Load(type, false);
```
sevoku:
```suggestion
			return Load(type, true);
```
Shouldn't this be `true` by default?

iainx: No...
false means it's not a guest, and therefore shouldn't do the full initialization

iainx: Yes, it should be false for this. It means "Do the full initialization"

So when Gtk is initialised through gtk.init, then Xwt.Gtk is the guest, and so doesn't need to do the initialization again (this isn't really an issue because Gtk allows Gtk.Init to be called multiple times)

When Cocoa is initialised through NSApplication.Initialize then Xwt.Mac is the guest, so doesn't need to do initialization

But when Xwt.Mac is running inside Gtk, then it is not the guest and so it needs to still do the initialization

iainx:
Separating the guest concept from whether or not to initialise the toolkit makes things a lot clearer

iainx:
Bump to test VSMac build here: https://github.com/xamarin/vsmac/pull/675

PR #901, opened 2018-12-06
[a11y][WPF] AutomationPeer customization support for TreeView and ListView

DmitriyKirakosyan:
@sevoku Please review.

### DIFF
--- a/Xwt.Gtk/Xwt.GtkBackend/AccessibleBackend.cs
+++ b/Xwt.Gtk/Xwt.GtkBackend/AccessibleBackend.cs
@@ -24,6 +24,8 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using Xwt.Accessibility;
 using Xwt.Backends;
 
@@ -190,6 +192,12 @@ namespace Xwt.GtkBackend
 		public virtual void RemoveAllChildren ()
 		{
 			// TODO
+		}
+
+		public virtual IEnumerable<object> GetChildren ()
+		{
+			// TODO
+			return Enumerable.Empty<object>();
 		}
 
 		public void DisableEvent (object eventId)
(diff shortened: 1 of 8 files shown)


DmitriyKirakosyan:
This changes allow to setup AutomationPeer for TreeView and its children (TreeViewItem).
Client doesn't need to access to TreeViewItem in order to create automation peer for it. The idea here is that TreeView's automation peer has a specific method for its children peers creation:

https://github.com/mono/xwt/pull/901/files#diff-3d9a78c59e4d79eb6619e47a53e3b0a7R507

that can be implemented in client.

DmitriyKirakosyan:
cc @BretJohnson ^

Xwt.XamMac/Xwt.Mac/AccessibleBackend.cs
```diff
@@ -240,6 +241,16 @@ public void RemoveAllChildren ()
 			accessible.AccessibilityChildren = null;
 		}
 
+		public IEnumerable<object> GetChildren ()
+		{
+			var accessible = widget as INSAccessibility;
+			if (accessible == null) {
+				return new List<object> ();
```
sevoku: `return Enumerable.Empty<object>();`

Xwt/Xwt.Accessibility/Accessible.cs
```diff
@@ -289,6 +295,11 @@ public void RemoveAllChildren ()
 		{
 		}
 
+		public IEnumerable<object> GetChildren ()
+		{
+			return new List<object>();
```
sevoku: `return Enumerable.Empty<object>();`

Xwt.Gtk/Xwt.GtkBackend/AccessibleBackend.cs
```diff
@@ -192,6 +193,12 @@ public virtual void RemoveAllChildren ()
 			// TODO
 		}
 
+		public virtual IEnumerable<object> GetChildren ()
+		{
+			// TODO
+			return new List<object> ();
```
sevoku: `return Enumerable.Empty<object>();`

DmitriyKirakosyan:
@sevoku ^

Xwt/Xwt/TreeNavigator.cs
```diff
@@ -115,6 +115,11 @@ public TreeNavigator AddChild ()
 			pos = backend.AddChild (pos);
 			return this;
 		}
+
+		public int GetChildrenCount ()
```
sevoku: This doesn't seem to be related, can we move this into a separate commit pls?

DmitriyKirakosyan: Removed. I think it is not needed anymore.

PR #902, opened 2018-12-08
[Xwt] Add some missing Key constants

iainx:
Adds some missing key constants


Xwt/Xwt/Key.cs
```diff
@@ -159,9 +159,19 @@ public enum Key
 		HyperRight = 0xffee,
 
 		Caret = 0xfe52,
+		Exclaimation = 0x21,
```
Therzok:
```suggestion
		Exclamation = 0x21,
```

Issue #817, opened 2018-03-26
Provide a way to determine if the dialog was cancelled in the FileSelector and FolderSelector


mrward:
The current behaviour of the FileSelector is that if the user browses to a different folder, then cancels the dialog, then next time the FileSelector is opened the folder browsed to is selected. Since there is no event fired that indicates that the dialog was cancelled it does not seem to be possible to reset the CurrentFolder property.

Ideas:
  Add a Canceled event.
  Add a `PreserveFolder` property.
  Add a BeforeBrowse event - fired before the dialog is opened.

stoiandan:
I'd like to take this. But I'm new here, will first need to setup my env, and get things working :)
So add a Canceled event in the FileSelector class?

mrward:
The original problem I had resulted in a workaround being added to [monodevelop](https://github.com/mono/monodevelop/commit/4d5934387c35353253bd9e1889c65156d3a83f27#diff-145c6fb4c19a4d10666d2151593b8a4e) since the XWT FileSelector did not work quite the same as the Gtk one available from MonoDevelop.Components.

The FileSelector has a FileChanged event but there is no way to know if the dialog was cancelled.

Another idea would be to have a Canceled property on the FileSelector instead of an event.

sevoku:
I'd prefer events with `WidgetEventArgs`:

**`event EventHandler<WidgetEventArgs> Browse`**
* fired before browse, can be overridden/suppressed by `WidgetEventArgs.Handled`
* let's call it `Browse` to follow the Xwt event naming pattern without prefixes (with virtual `OnBrowse()`)

**`event EventHandler<SelectorDialogClosedEventArgs> SelectorClosed`**
* fired after the dialog has been closed
* update `FileSelector.FilePath` and `FileSelector.CurrentFolder` only if not `SelectorDialogClosedEventArgs.Handled`

With the following EventArgs:
```c#
public class SelectorDialogClosedEventArgs : WidgetEventArgs
{
	// true if canceled
	public bool Canceled { get; }
}

public class FileSelectorDialogClosedEventArgs : SelectorDialogClosedEventArgs
{
	// picked folder
	public string SelectedFolder { get; }

	// picked path (absolute, inside picked folder)
	public string SelectedFilePath { get; }
}
```
This would allow us to add the same `Canceled` functionality to other selectors (FontSelector, ColorSelector, etc. + a common selector interface) later.

slluis:
[...]

PR #905, opened 2018-12-20
#817 prevent currentFolder to be overridden by canceled dialog

stoiandan:


stoiandan:
I've noted that the `currentFolder `property is being set twice, I don't see any reason for that, so I've removed it.

PR #913, opened 2019-01-24
[Gtk] Fix the focus order of widgets inside a Box

sevoku:
We don't use HBox/VBox containers but let the frontend calculate
the widget allocation. This seems to confuse the Gtk focus order
calculation, especially if the widgets have additional alignemnt
requirements (vertical in HBox and horizontal in VBox).

Fixes VSTS #752833

### DIFF
--- a/Xwt.Gtk/Xwt.GtkBackend/BoxBackend.cs
+++ b/Xwt.Gtk/Xwt.GtkBackend/BoxBackend.cs
@@ -94,7 +94,12 @@ namespace Xwt.GtkBackend
 			Add (newWidget);
 			children [newWidget] = r;
 		}
-		
+
+		void UpdateFocusChain (Orientation orientation)
+		{
+			FocusChain = children.OrderBy ((arg) => orientation == Orientation.Horizontal ? arg.Value.Rect.X : arg.Value.Rect.Y).Select (arg => arg.Key).ToArray ();
+		}
+
 		public bool SetAllocation (Gtk.Widget w, Rectangle rect)
 		{
 			WidgetData r;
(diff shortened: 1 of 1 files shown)


Xwt.Gtk/Xwt.GtkBackend/BoxBackend.cs
```diff
@@ -94,14 +94,20 @@ public void ReplaceChild (Gtk.Widget oldWidget, Gtk.Widget newWidget)
 			Add (newWidget);
 			children [newWidget] = r;
 		}
-		
+
+		void UpdateFocusChain (Orientation orientation)
+		{
+			FocusChain = children.OrderBy ((arg) => orientation == Orientation.Horizontal ? arg.Value.Rect.X : arg.Value.Rect.Y).Select (arg => arg.Key).ToArray ();
```
Therzok: Don't like the use of LINQ here, would prefer creating a new array and using an EqualityComparer which looks into the dictionary for the WidgetData.

Therzok: Some pseudocode from phone:
```
class EqComparer : IComparer<Gtk.Widget>
{
    public EqComparer (Dictionary<Gtk.Widget, WidgetData> map, Orientation orientation)
    {
        // store the values
    }
    
    public int Compare (a, b)
  {
    int left, right;
    if (orientation == horizontal) {
        left = map[a].x;
        right = map[b].x;
    } else {
        left = map[a].y;
        right = map[b].y;
    }
    return left - right;
  }
}

```

Issue #795, opened 2018-03-07
CheckBoxCellView.Toggled missleading


mkrueger:
Toggled is called before the state change of the list - so it's actual not a Toggled it's a Toggling event.

If someone wants to update the state of an ok button for example the state after the toggling is important and the name Toggled is missleading.

PR #914, opened 2019-01-25
[Gtk] Adds EditingFinished event to cell renderers

netonjm:
This PR allows handle the event when the edition has finished. It's very useful to get real values from the ListStore.

I didn't implemented the Cocoa side because there are no implementation, but this can be added in the future.

values


Therzok:
Should we pass over the old/new value in the event args?

netonjm:
This happens when the value is set in your model, then you can check it here

Therzok:
The reason the event args are used is so event handlers do not have to query.

In example, looking up the value in a store or looking up which row changed can be costly, and if 2 event handlers are subscribed, we do potentially expensive lookups instead of doing a cheap one at eventargs construction timem

netonjm:
It's a good reason to include it 👍

sevoku:
@netonjm are you going to add the event args?

This should fix #795 and we could also fix mono/monodevelop#4075

netonjm:
@sevoku @Therzok Fixed and ready!

Xwt/Xwt.Backends/IComboBoxCellViewFrontend.cs
```diff
@@ -27,9 +27,22 @@
 
 namespace Xwt.Backends
 {
+	public class EditableCellViewArgs : EventArgs
```
sevoku: This should be `Xwt.CellEditingFinishedArgs` (and separate file in Xwt folder and namespace) and probably even `Xwt.CellEditingFinishedArgs<T>` where T is the most convenient type (for CheckBox it would be `CheckBoxState` and not `bool` for example).

netonjm: Fixed!

Xwt/Xwt.Backends/IToggleCellViewFrontend.cs
```diff
@@ -27,7 +27,7 @@
 namespace Xwt.Backends
 {
 
-	public interface IToggleCellViewFrontend : ICellViewFrontend, IEditableCellViewFrontend
+	public interface IToggleCellViewFrontend : ICellViewFrontend, IEditableCellViewFrontend<CheckBoxState>
```
sevoku: Toggle has no mixed state and should be `IEditableCellViewFrontend<bool>`, probably `IToggleCellViewFrontend` should not be based on `IEditableCellViewFrontend` but only the implementing frontend classes (otherwise it would be ambiguous for check boxes).

netonjm: Fixed

sevoku:
LGTM now! cc @slluis for the last API change

Therzok:
We should rebase and merge this. Stumbled upon https://github.com/mono/monodevelop/issues/4075

PR #918, opened 2019-02-01
[build] Add a Directory.Build.targets

alanmcgovern:
We have a Directory.Build.props so we should also add a .targets.

This is important because MSBuild searches upwards for both of
those files, so if we have a props and do not also have a targets
then it will be possible that we'll load up someone else's targets
*and not* also load up their props.

This affected the designer's build as Xwt skipped our props but took
our targets, and our targets depended on our props  :p


alanmcgovern:
YOLO!

PR #923, opened 2019-02-11
Fix build - Track IAccessibleBackend API changes

sevoku:


Xwt/Xwt.Accessibility/XwtAccessibleBackend.cs
```diff
@@ -211,5 +212,10 @@ public void EnableEvent (object eventId)
 		{
 			nativeBackend.EnableEvent (eventId);
 		}
+
+		public IEnumerable<object> GetChildren()
```
Therzok: Why is this done via IEnumerable?

sevoku: It has been introduced here: https://github.com/mono/xwt/pull/901/files#diff-798533a5d4653f71651d0fde0885384bR223
I think it's ok in this case because of API consistency (basically all `Children` properties are IEnumerable, like `Table.Children`,`Canvas.Children`, `Box.Children`) and WPF has only enumerables too. Only on Mac it's an array and Gtk doesn't support this at all.

PR #929, opened 2019-03-28
[Mac] Avoid touching the Delegate of a wrapped window

sevoku:
When wrapping an NSWindow with an existing delegate, we
should not subscribe to any managed events and trigger
the default delegate creation.

Fixes VSTS #831134

### DIFF
--- a/Xwt.XamMac/Xwt.Mac/WindowFrameBackend.cs
+++ b/Xwt.XamMac/Xwt.Mac/WindowFrameBackend.cs
@@ -37,6 +37,7 @@ namespace Xwt.Mac
 	{
 		IWindowFrameEventSink eventSink;
 		WindowFrame frontend;
+		bool hasExternalDelegate;
 
 		public WindowFrameBackend ()
 		{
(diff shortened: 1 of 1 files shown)


Xwt.XamMac/Xwt.Mac/WindowFrameBackend.cs
```diff
@@ -164,32 +164,37 @@ public void Present ()
...
 				case WindowFrameEvent.Shown:
 					EnableVisibilityEvent (@event);
 					break;
 				case WindowFrameEvent.CloseRequested:
-					Window.WindowShouldClose = OnShouldClose;
+					// NOTE: this works only if the wrapped window has no delegate
+					if (Window.Delegate == null && Window.WeakDelegate == null) // don't replace existing delegates
+						Window.WindowShouldClose = OnShouldClose;
+					else
+						return; // skip eventEnabled update to avoid touching WindowShouldClose when disabling events
```
Therzok: Should we explicitly throw here?

sevoku: I'd say no, this might break client logic and catching when subscribing to events is really ugly (but would be required then). I think it's ok if the event just wont ever fire. It's a very specific scenario.

sevoku: oh, but this needs another fix...

PR #931, opened 2019-04-01
[GTK] Close dialog on EndLoop to be consistent with XamMac and WPF

daserge:
https://github.com/mono/xwt/blob/master/Xwt.XamMac/Xwt.Mac/DialogBackend.cs#L181
https://github.com/mono/xwt/blob/master/Xwt.WPF/Xwt.WPFBackend/DialogBackend.cs#L166


Therzok:
LGTM. @sevoku for final sign off?

PR #938, opened 2019-05-22
[WPF] Respect TreeView column Expands flag

daserge:
This will automatically expand the columns marked with Expands flag when their content changes.
Shift cell views column index and global cell indexes following the removed TreeView column.
^ This also fixes ArgumentOutOfRangeException in TreeViewBackend->GetCellBounds

### DIFF
--- a/Xwt.WPF/Xwt.WPFBackend/ExTreeViewItem.cs
+++ b/Xwt.WPF/Xwt.WPFBackend/ExTreeViewItem.cs
@@ -48,7 +48,6 @@ namespace Xwt.WPFBackend
 		{
 			Loaded += OnLoaded;
 			HorizontalContentAlignment = HorizontalAlignment.Stretch;
-			
 		}
 
 		public ExTreeViewItem (ExTreeView view)
--- a/Xwt.WPF/Xwt.WPFBackend/ExTreeViewItem.cs
+++ b/Xwt.WPF/Xwt.WPFBackend/ExTreeViewItem.cs
@@ -48,6 +48,7 @@ namespace Xwt.WPFBackend
 		{
 			Loaded += OnLoaded;
 			HorizontalContentAlignment = HorizontalAlignment.Stretch;
+			
 		}
 
 		public ExTreeViewItem (ExTreeView view)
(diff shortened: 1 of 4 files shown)


Xwt.WPF/Xwt.WPFBackend/ExTreeViewItem.cs
```diff
@@ -226,6 +226,11 @@ private void OnLoaded (object sender, RoutedEventArgs routedEventArgs)
 			}
 		}
 
+		public void ExpandColumnWidth (int columnIndex, double toWidth)
```
sevoku:
```suggestion
		public void SetColumnWidth (int columnIndex, double toWidth)
```
and does this need to be `public`?

Xwt.WPF/Xwt.WPFBackend/TreeViewBackend.cs
```diff
@@ -234,6 +235,59 @@ public void UnselectAll ()
...
+						if (cell_bounds.Width > maxColumnViewWidth)
+							maxColumnViewWidth = cell_bounds.Width;
+					}
+
+					totalColumnWidth += maxColumnViewWidth;
+				}
+
+				treeViewItem.ExpandColumnWidth (changedColumnIndex, totalColumnWidth);
+			} catch (Exception ex) {
+				Console.WriteLine ($"[TreeViewBackend_NodeChanged] Exception: {ex}");
```
sevoku: Is there a reason to swallow the exception here?

daserge:
`GlobalCellIndex` is for sequential index of a CellView.

Xwt/Xwt/ITreeDataSource.cs
```diff
@@ -68,6 +72,8 @@ public TreeNodeEventArgs (TreePosition node, int childIndex)
 			get;
 			private set;
 		}
+
+		public int ChangedCellIndex { get; private set; }
```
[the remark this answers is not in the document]
daserge: Actually this is the field index here and not the column: https://github.com/mono/xwt/blob/af2540cb0b6a1c1e82c4f7ec8949565f6677e631/Xwt/Xwt/TreeNavigator.cs#L121
Maybe I should rename this to `FieldIndex` for clarity?

chkn:
@daserge @sevoku What's the status of this? We have a designer issue that would probably be resolved by this patch.

daserge:
@chkn I've encountered more issues with column width calculations which I didn't find a solution to yet.
The Designer issue has a workaround currently so I consider this as a lower priority task.
Let me reopen when its ready.

PR #940, opened 2019-05-29
Add an optional MessageDialog.RootWindow override

sevoku:
Allows the client to supply a custom dynamic RootWindow using MessageDialogDefaults.DefaultRootWindowOverride.

And adds some useful MessageDialog.Show* overloads.

NOTE: this is a quicj workaround to be less restricted to the static MessageService.RootWindow. The long term plan is to deprecate that property and let the toolkits decide which root window to use. But this need some more complex changes to the API.

### DIFF
--- a/Xwt/Xwt/MessageDialog.cs
+++ b/Xwt/Xwt/MessageDialog.cs
@@ -1,4 +1,4 @@
-// 
+﻿// 
 // MessageDialog.cs
 //  
 // Author:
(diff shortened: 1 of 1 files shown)


Xwt/Xwt/MessageDialog.cs
```diff
@@ -32,8 +32,20 @@ namespace Xwt
...
 	{
-		public static WindowFrame RootWindow { get; set; }
-		
+		private static WindowFrame rootWindow;
+
+		public static WindowFrame RootWindow
+		{
+			get {
+				if (Toolkit.CurrentEngine.Defaults.MessageDialog.DefaultRootWindowOverride != null)
+					return Toolkit.CurrentEngine.Defaults.MessageDialog.DefaultRootWindowOverride() ?? rootWindow;
```
Therzok: The API here is weird. Why do we want this as a `Func<WindowFrame>`?

sevoku: The problem is that if you show some dialog and a message using the message service, you need either to call an override with a custom parent window or set the RootWindow accordingly (and then back). It becomes even worse if you have different main windows (like VS for Mac has a main Shell and a GetToCode dialog). This also becomes a real problem if you consume some external API which is using Xwt and doesn't know your code. It won't know about your windows structure and always use RootWindow (like the Android SDK Manager in VS for Mac).

PR #941, opened 2019-05-30
[Mac] Fixed bug with Popover where the TextEntry is gray.

viniciusjarina:
* There is is a bug on AppKit where if you don't set the `Appearance` for you view, the `NSTextField` will be displayed with gray background

https://stackoverflow.com/questions/29074724/how-to-change-nstextfield-background-color-in-nspopover


viniciusjarina:
The appveyor has an old version of Xamarin.Mac, the build is fine.

sevoku:
This PR not only broke the builds, it also broke compatibility with any macOS version older than Mojave.

PR #945, opened 2019-07-09
[Mac] Use INSWindowDelegate instead of NSWindow events

sevoku:
Fixes VSTS #935131

### DIFF
--- a/Xwt.XamMac/Xwt.Mac/WindowBackend.cs
+++ b/Xwt.XamMac/Xwt.Mac/WindowBackend.cs
@@ -243,9 +243,7 @@ namespace Xwt.Mac
 		[Export ("windowWillClose:")]
 		new void WillClose (NSNotification notification)
 		{
-			if (eventsEnabled.HasFlag (WindowFrameEvent.Hidden)) {
-				OnHidden ();
-			}
+			OnHidden ();
 			OnClosed ();
 		}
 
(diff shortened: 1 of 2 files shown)


Xwt.XamMac/Xwt.Mac/WindowBackend.cs
```diff
@@ -211,33 +210,45 @@ void IBackend.EnableEvent (object eventId)
 			if (eventId is WindowFrameEvent) {
 				var @event = (WindowFrameEvent)eventId;
 				switch (@event) {
-				case WindowFrameEvent.BoundsChanged:
```
Therzok: Shown/Hidden uses observers and this uses selectors. Should we unify on one pattern? There's DidResizeNotification and DidMoveNotification.

Xwt.XamMac/Xwt.Mac/WindowBackend.cs
```diff
@@ -211,33 +210,45 @@ void IBackend.EnableEvent (object eventId)
...
-					EnableVisibilityEvent (@event);
-					break;
-				case WindowFrameEvent.CloseRequested:
-					WindowShouldClose = OnShouldClose;
-					break;
+					case WindowFrameEvent.Hidden:
+					case WindowFrameEvent.Shown:
+						if (!VisibilityEventsEnabled())
+						{
+							ContentView.AddObserver(this, HiddenProperty, NSKeyValueObservingOptions.New, IntPtr.Zero);
```
Therzok: Store the observer and dispose it on window dispose.

Xwt.XamMac/Xwt.Mac/WindowBackend.cs
sevoku: Using selectors with the delegate is cleaner imo, so I'd prefer keeping it. Unfortunately we can't change thjat for the observer cos we observe ContentView and not the window.

Xwt.XamMac/Xwt.Mac/WindowBackend.cs
sevoku: hhm, `AddObserver` is `void` (it's not like NotificationManager) and we remove it here: https://github.com/mono/xwt/blob/512728f98e8390de4919b85ee7177a8e1b35994b/Xwt.XamMac/Xwt.Mac/WindowBackend.cs#L477

Therzok: Ah, I see. Good.

PR #951, opened 2019-07-24
[Mac] Preserve the window location when resizing vertically

sevoku:
Since the y-axis is inverted on Mac, we need to recalculate
the Y position to keep the window in place when it's being
resized vertically.


sevoku:
Simple demo:

```csharp
Dialog d = new Dialog ();
d.Title = "This is a dialog";
var dbox = new VBox();

var expander = new Expander();
expander.Label = "A title here";
var content = new VBox();
content.PackStart(new Label() { Text = "Label 1" });
content.PackStart(new Button() { Label = "Button 2" });
expander.Content = content;
				
dbox.PackStart(expander);
dbox.PackStart(new Label("This is shown below the expander"));

d.Content = dbox;
				
d.Buttons.Add (new DialogButton (Command.Cancel));
d.Buttons.Add (new DialogButton (Command.Ok));
d.DefaultCommand = Command.Ok;

var r = d.Run ();
d.Dispose ();
```

Without the fix:
Window Resize Position broken

Fixed:
Window Resize Position fixed

PR #956, opened 2019-07-30
[Mac] Fix Popover appearance on Mac.

viniciusjarina:
The `NativeChild.EffectiveAppearance` is `Aqua` in most cases since the NativeChild was just created. Use the EffectiveAppearance from the reference view instead.


viniciusjarina:
@monojenkins backport d16-3

PR #958, opened 2019-07-31
Improve select font dialog

sevoku:


mkrueger:
Added to PR
https://github.com/mono/xwt/pull/960

PR #959, opened 2019-08-02
[WPF][XamMac][GtkMac] Enable formatted text in Buttons

mcumming:
Adding MenuButton support for formatted text to be used in MenuItems and the selected item. 

Backends:
- [x] WPF
- [x] GtkMac
- [x] XamMac

### DIFF
--- a/Xwt.Gtk/Xwt.GtkBackend/ButtonBackend.cs
+++ b/Xwt.Gtk/Xwt.GtkBackend/ButtonBackend.cs
@@ -197,6 +197,28 @@ namespace Xwt.GtkBackend
 				Widget.Label = null;
 		}
 		
+		private void SetFormattedContent (FormattedText label, ContentPosition position)
+		{
+			if (label != null && label.Text.Length == 0)
+				label = null;
+
+			Button b = (Button)Frontend;
+			if (label != null && image.Backend == null && b.Type == ButtonType.Normal) {
+				labelWidget = new Gtk.Label ();
+				label.ApplyToLabel (labelWidget);
+				Widget.Image = labelWidget;
+				return;
+			}
+
+			if (b.Type == ButtonType.Disclosure) {
+				return;
+			}
+
+			if (label != null && labelWidget != null) {
+				label.ApplyToLabel (labelWidget);
+			}
+		}
+
 		public void SetButtonStyle (ButtonStyle style)
 		{
 			switch (style) {
@@ -223,10 +245,8 @@ namespace Xwt.GtkBackend
 
 		public void SetFormattedText (FormattedText text)
 		{
-			if (labelWidget != null) {
-				labelWidget.Text = text.Text;
-				text.ApplyToLabel (labelWidget);
-			}
+			SetContent (text.Text, Widget.UseUnderline, this.image, ContentPosition.Center);
+			SetFormattedContent (text, ContentPosition.Center);
 		}
 
 		public override void EnableEvent (object eventId)
--- a/Xwt.Gtk/Xwt.GtkBackend/ButtonBackend.cs
+++ b/Xwt.Gtk/Xwt.GtkBackend/ButtonBackend.cs
@@ -316,6 +316,7 @@ namespace Xwt.GtkBackend
 			if (disposing && labelWidget != null) {
 				labelWidget.Realized -= HandleStyleUpdate;
 				labelWidget.StyleSet -= HandleStyleUpdate;
+				labelWidget.Destroy ();
 				labelWidget = null;
 			}
 			base.Dispose (disposing);
--- a/Xwt.Gtk/Xwt.GtkBackend/ButtonBackend.cs
+++ b/Xwt.Gtk/Xwt.GtkBackend/ButtonBackend.cs
@@ -197,25 +197,32 @@ namespace Xwt.GtkBackend
 				Widget.Label = null;
 		}
 		
+		FormattedText formattedText = null;
 		private void SetFormattedContent (FormattedText label, ContentPosition position)
 		{
 			if (label != null && label.Text.Length == 0)
-				label = null;
+				return;
 
 			Button b = (Button)Frontend;
-			if (label != null && image.Backend == null && b.Type == ButtonType.Normal) {
-				labelWidget = new Gtk.Label ();
-				label.ApplyToLabel (labelWidget);
+			if (labelWidget == null) {
+				labelWidget = new Gtk.Label (label.Text);
+				Widget.Label = null;
 				Widget.Image = labelWidget;
-				return;
 			}
 
-			if (b.Type == ButtonType.Disclosure) {
-				return;
-			}
+			formattedText = label;
+			labelWidget.ApplyFormattedText (formattedText);
+			labelWidget.ShowAll ();
+			labelWidget.Realized += HandleStyleUpdate;
+			labelWidget.StyleSet += HandleStyleUpdate;
+		}
 
-			if (label != null && labelWidget != null) {
-				label.ApplyToLabel (labelWidget);
+		void HandleStyleUpdate (object sender, EventArgs e)
+		{
+			// force text update with updated link color
+			if (labelWidget.IsRealized && formattedText != null) {
+				labelWidget.ApplyFormattedText (formattedText);
+				labelWidget.ShowAll ();
 			}
 		}
 
@@ -299,6 +306,15 @@ namespace Xwt.GtkBackend
 		{
 			Widget.Child.SizeAllocate (args.Allocation);
 			args.RetVal = true;
+		}
+
+		protected override void Dispose (bool disposing)
+		{
+			if (labelWidget != null) {
+				labelWidget.Realized -= HandleStyleUpdate;
+				labelWidget.StyleSet -= HandleStyleUpdate;
+			}
+			base.Dispose (disposing);
 		}
 	}
 }
--- a/Xwt.Gtk/Xwt.GtkBackend/ButtonBackend.cs
+++ b/Xwt.Gtk/Xwt.GtkBackend/ButtonBackend.cs
@@ -220,7 +220,15 @@ namespace Xwt.GtkBackend
 			Button b = (Button) Frontend;
 			SetContent (b.Label, b.UseMnemonic, image, b.ImagePosition);
 		}
-		
+
+		public void SetFormattedText (FormattedText text)
+		{
+			if (labelWidget != null) {
+				labelWidget.Text = text.Text;
+				text.ApplyToLabel (labelWidget);
+			}
+		}
+
 		public override void EnableEvent (object eventId)
 		{
 			base.EnableEvent (eventId);
--- a/Xwt.Gtk/Xwt.GtkBackend/ButtonBackend.cs
+++ b/Xwt.Gtk/Xwt.GtkBackend/ButtonBackend.cs
@@ -130,6 +130,8 @@ namespace Xwt.GtkBackend
 			if (image.Backend != null)
 				imageWidget = new ImageBox (ApplicationContext, image.WithDefaultSize (Gtk.IconSize.Button));
 
+			labelWidget.Realized -= HandleStyleUpdate;
+			labelWidget.StyleSet -= HandleStyleUpdate;
 			labelWidget = null;
 
 			if (label != null && imageWidget == null) {
@@ -209,6 +211,8 @@ namespace Xwt.GtkBackend
 				Widget.Label = null;
 				Widget.Image = labelWidget;
 			}
+			labelWidget.Realized -= HandleStyleUpdate;
+			labelWidget.StyleSet -= HandleStyleUpdate;
 
 			formattedText = label;
 			labelWidget.ApplyFormattedText (formattedText);
@@ -252,7 +256,7 @@ namespace Xwt.GtkBackend
 
 		public void SetFormattedText (FormattedText text)
 		{
-			SetContent (text.Text, Widget.UseUnderline, this.image, ContentPosition.Center);
+			SetContent (text?.Text, Widget.UseUnderline, this.image, ContentPosition.Center);
 			SetFormattedContent (text, ContentPosition.Center);
 		}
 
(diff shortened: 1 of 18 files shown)


Xwt.WPF/Xwt.WPFBackend/ButtonBackend.cs
```diff
@@ -100,39 +101,50 @@ protected ButtonBackend (ButtonBase impl)
...
-					Button.Content = accessText;
-				else
-					Button.Content = accessText.Text.Replace ("_", "__");
-			else {
+			accessText.Text = label ?? string.Empty;
+			var mnemonicValue = useMnemonic ? accessText.Text : accessText.Text.Replace ("_", "__");
+
+			if (image.IsNull) {
+				var formattedText = FormattedText.FromMarkup (mnemonicValue);
+				if (mnemonicValue.CompareTo (formattedText.Text) != 0) {
```
Therzok:
```suggestion
				if (!mnemonicValue.Equals (formattedText.Text)) {
```

Xwt.WPF/Xwt.WPFBackend/ButtonBackend.cs
```diff
@@ -62,35 +62,36 @@ protected ButtonBackend (ButtonBase impl)
 			get { return (IButtonEventSink)base.EventSink; }
 		}
 
-		public void SetButtonStyle (ButtonStyle style) {
```
sevoku: Please avoid unrelated formatting changes

Xwt.WPF/Xwt.WPFBackend/MenuItemBackend.cs
```diff
@@ -35,11 +35,12 @@
...
 using SWMI = System.Windows.Media.Imaging;
 using Xwt.Backends;
-
+using Xwt.WPFBackend.Utilities;
+using WindowsOrientation = System.Windows.Controls.Orientation;
 
 namespace Xwt.WPFBackend
 {
-	public class MenuItemBackend : Backend, IMenuItemBackend
+	public class MenuItemBackend : WidgetBackend, IMenuItemBackend
```
sevoku: Why is this needed?

Xwt.Gtk/Xwt.GtkBackend/MenuItemBackend.cs
```diff
@@ -50,6 +50,11 @@ public MenuItemBackend (Gtk.MenuItem item)
 		{
 			this.item = item;
 			label = (Gtk.Label) item.Child;
+			if (label != null) {
+				label.Realized += HandleStyleUpdate;
+				label.StyleSet += HandleStyleUpdate;
```
netonjm: this should be unsubscribed on disposing ?

Xwt/Xwt.Backends/IButtonBackend.cs
```diff
@@ -36,6 +36,12 @@ public interface IButtonBackend: IWidgetBackend
 		void SetButtonStyle (ButtonStyle style);
 		void SetButtonType (ButtonType type);
 		void SetContent (string label, bool useMnemonic, ImageDescription image, ContentPosition position);
+
+		/// <summary>
+		/// Sets the formatted text display for this menu item.
+		/// </summary>
+		/// <param name="text">The text to display with additional markup for formatting <see cref="Xwt.FormattedText"/></param>
+		void SetFormattedText (FormattedText text);
```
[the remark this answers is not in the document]
mcumming: I don't think it would make much of a difference on the implementation of the backend either way. I think what would make it easier, was if all the labels were converted to FormattedText objects then applied. But you would need a way to determine of the FormattedText did not in fact contain any formatting so that you could short curcuit the label creation and styling for plain text.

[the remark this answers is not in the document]
sevoku: well, it'd let us avoid double-init here: https://github.com/mono/xwt/pull/959/files#diff-75a1dc4c07ad775d27eb0deadeee862fR255. Whether FormattedText contains formatting or not can be determined by `FormattedText.Attributes.Count`.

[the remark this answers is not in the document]
mcumming: So my reasoning for setting it up that way was that if you change the `Label` of the button, it calls `SetContent`, which recreates the `labelWidget` internally. So setting the `Markup` property it should do all the things that setting the `Label` does, plus applying the proper formatting. This is why there is a check to see if the `labelWidget` already exists in `SetFormattedContent` instead of just recreating it every time. In this way, `SetFormattedText` uses `SetContent` to configure everything properly, and then `SetFormattedContent` extends that to add the proper formatting.

Xwt.Gtk/Xwt.GtkBackend/ButtonBackend.cs
```diff
@@ -220,7 +249,13 @@ public void SetButtonType (ButtonType type)
 			Button b = (Button) Frontend;
 			SetContent (b.Label, b.UseMnemonic, image, b.ImagePosition);
 		}
-		
+
+		public void SetFormattedText (FormattedText text)
+		{
+			SetContent (text.Text, Widget.UseUnderline, this.image, ContentPosition.Center);
```
netonjm: possible nre in text

Xwt.Gtk/Xwt.GtkBackend/MenuItemBackend.cs
```diff
@@ -168,7 +168,25 @@ void ImageMenuItemStateChanged (object o, Gtk.StateChangedArgs args)
 				}
 			}
 		}
-		
+
+		FormattedText formattedText = null;
+		public void SetFormattedText (FormattedText text)
+		{
+			label.Text = text.Text;
```
netonjm: possible nre in text

Xwt.Gtk/Xwt.GtkBackend/ButtonBackend.cs
```diff
@@ -272,6 +307,15 @@ void HandleSizeAllocated (object o, Gtk.SizeAllocatedArgs args)
 			Widget.Child.SizeAllocate (args.Allocation);
 			args.RetVal = true;
 		}
+
+		protected override void Dispose (bool disposing)
+		{
+			if (labelWidget != null) {
+				labelWidget.Realized -= HandleStyleUpdate;
```
sevoku: `labelWidget` is created either in `SetFormattedText()` (events subscribed) or in `SetContent ()` (events not subscribed), so this will thrwo NRE for regular text (SetFormattedText() not called).

Xwt.Gtk/Xwt.GtkBackend/ButtonBackend.cs
```diff
@@ -272,6 +307,15 @@ void HandleSizeAllocated (object o, Gtk.SizeAllocatedArgs args)
 			Widget.Child.SizeAllocate (args.Allocation);
 			args.RetVal = true;
 		}
+
+		protected override void Dispose (bool disposing)
+		{
+			if (labelWidget != null) {
```
sevoku:
```suggestion
			if (disposing && labelWidget != null) {
```
and set `labelWidget = null` at the end of the block to not NRE on double disposing.

Xwt.Gtk/Xwt.GtkBackend/ButtonBackend.cs
```diff
@@ -197,6 +197,35 @@ public void SetContent (string label, bool useMnemonic, ImageDescription image,
...
+			if (labelWidget == null) {
+				labelWidget = new Gtk.Label (label.Text);
+				Widget.Label = null;
+				Widget.Image = labelWidget;
+			}
+
+			formattedText = label;
+			labelWidget.ApplyFormattedText (formattedText);
+			labelWidget.ShowAll ();
+			labelWidget.Realized += HandleStyleUpdate;
```
sevoku: Will subscribe on every call, what if the Markup property get's updated?

Xwt.Gtk/Xwt.GtkBackend/MenuItemBackend.cs
```diff
@@ -168,7 +168,25 @@ void ImageMenuItemStateChanged (object o, Gtk.StateChangedArgs args)
...
+			formattedText = text;
+			label.Realized += HandleStyleUpdate;
```
sevoku: same as in button, multiple subscription on every Markup change.

Xwt.Gtk/Xwt.GtkBackend/MenuItemBackend.cs
```diff
@@ -249,6 +267,14 @@ void HandleItemActivated (object sender, EventArgs e)
 				context.InvokeUserCode (eventSink.OnClicked);
 			}
 		}
+
+		public void Dispose ()
+		{
+			if (label != null) {
```
sevoku: label never set to null, possible NRE on double dispose.

Xwt.XamMac/Xwt.Mac/ButtonBackend.cs
```diff
@@ -153,8 +153,12 @@ public void SetButtonType (ButtonType type)
 			}
 		}
 
+		public void SetFormattedText (FormattedText text)
+		{
+			Widget.Title = text.Text;
```
sevoku: Is it necessary to set the Title here? It will be overwritten by AttributedTitle in the next line.

mcumming: Actually, this never formats the selected item in the button properly anyway, so I can remove that line.

Xwt.Gtk/Xwt.GtkBackend/ButtonBackend.cs
```diff
@@ -197,6 +199,37 @@ public void SetContent (string label, bool useMnemonic, ImageDescription image,
 				Widget.Label = null;
 		}
 		
+		FormattedText formattedText = null;
+		private void SetFormattedContent (FormattedText label, ContentPosition position)
+		{
+			if (label != null && label.Text.Length == 0)
```
Therzok: Don't we want to clear the content when label is null or it has no text?

This seems to exit when label is not null and the text is empty. Then when `labelWidget.ApplyFormattedText (formattedText)` is called, it's going to NRE inside, because formattedText is null.

Xwt.Gtk/Xwt.GtkBackend/ButtonBackend.cs
```diff
@@ -197,6 +199,37 @@ public void SetContent (string label, bool useMnemonic, ImageDescription image,
...
+				labelWidget = new Gtk.Label (label.Text);
+				Widget.Label = null;
+				Widget.Image = labelWidget;
+			}
+			labelWidget.Realized -= HandleStyleUpdate;
+			labelWidget.StyleSet -= HandleStyleUpdate;
+
+			formattedText = label;
+			labelWidget.ApplyFormattedText (formattedText);
+			labelWidget.ShowAll ();
```
Therzok: Is this ShowAll needed?

Xwt.Gtk/Xwt.GtkBackend/MenuItemBackend.cs
```diff
@@ -168,7 +168,28 @@ void ImageMenuItemStateChanged (object o, Gtk.StateChangedArgs args)
...
-		
+
+		FormattedText formattedText = null;
+		public void SetFormattedText (FormattedText text)
+		{
+			label.Text = text?.Text;
+			formattedText = text;
+			label.Realized -= HandleStyleUpdate;
+			label.StyleSet -= HandleStyleUpdate;
+			label.ApplyFormattedText(text);
```
Therzok: This will NRE inside ApplyFormattedText as `text` can be null.

Xwt.Gtk/Xwt.GtkBackend/ButtonBackend.cs
```diff
@@ -130,6 +130,8 @@ public void SetContent (string label, bool useMnemonic, ImageDescription image,
 			if (image.Backend != null)
 				imageWidget = new ImageBox (ApplicationContext, image.WithDefaultSize (Gtk.IconSize.Button));
 
+			labelWidget.Realized -= HandleStyleUpdate;
```
sevoku: This NREs on first call

Xwt.Gtk/Xwt.GtkBackend/ButtonBackend.cs
sevoku: yep, good point

Xwt.Gtk/Xwt.GtkBackend/ButtonBackend.cs
sevoku: shouldn't be needed

Xwt.XamMac/Xwt.Mac/MenuItemBackend.cs
```diff
@@ -149,6 +149,11 @@ public bool Checked
 			}
 		}
 
+		public void SetFormattedText (FormattedText text)
+		{
+			item.AttributedTitle = text.ToAttributedString ();
```
sevoku: Once set, it'll be always used, it's impossible to set an unformatted label afterwards.

Xwt.Gtk/Xwt.GtkBackend/MenuBackend.cs
```diff
@@ -109,6 +109,7 @@ public void RemoveItem (IMenuItemBackend menuItem)
 		{
 			Gtk.MenuItem item = ((MenuItemBackend)menuItem).MenuItem;
 			menu.Remove (item);
+			item.Dispose ();
```
Therzok: Gtk code usually calls Destroy, not Dispose.

Xwt/Xwt/MenuItem.cs
```diff
@@ -227,6 +243,12 @@ protected virtual void OnClicked (EventArgs e)
 				base.BackendHost.OnAfterEventRemove (MenuItemEvent.Clicked, clicked);
 			}
 		}
+
+		protected override void Dispose (bool release_all)
+		{
+			Backend.Dispose ();
```
Therzok: We should only Dispose when `release_all` is true, otherwise we're going to be doing the following:

```
MenuItem.Finalizer() -> MacMenuItem.Dispose() -> MacMenuItem.Dispose(true)
```

Which will end up throwing an AppKit inconsistency exception for not being on the UI Thread.

PR #961, opened 2019-08-06
[Windows] Fix AlertDialogBackend on Windows

dkandrov:
The result of working WPF backend was wrong and looked bad. I changed it to use Text as Title and SecondaryText as a base message of resulting dialog.

The system message box as backend:
image

The custom dialog as backend:
image

### DIFF
--- a/Xwt.WPF/Xwt.WPFBackend/AlertDialogBackend.cs
+++ b/Xwt.WPF/Xwt.WPFBackend/AlertDialogBackend.cs
@@ -61,17 +61,13 @@ namespace Xwt.WPFBackend
 				// Use a system message box
 				if (message.SecondaryText == null)
 					message.SecondaryText = String.Empty;
-				else {
-					message.Text = message.Text + "\r\n\r\n" + message.SecondaryText;
-					message.SecondaryText = String.Empty;
-				}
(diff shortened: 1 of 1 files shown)


slluis:
This looks wrong to me. The title of the window should be the application name. Also, the secondary text is optional, so if it is not provided the dialog will be empty.

sevoku:
Right, I 've been too fast with mergeing, let's revert. I think we need to introduce an optional Title parameter, since on Windows there is no well defined application name.

dkandrov:
@sevoku I'll try to implement it.

sevoku:
Reverted...

@dkandrov sry for the confusion, you'll need to rebase your changes on master, otherwise GH will think that it's already merged.
Let's add [`MessageDescription.Title`](https://github.com/mono/xwt/blob/master/Xwt/Xwt/MessageDialog.cs#L240) and maybe some common overloads in `MessageDialog`. In Gtk the Title can be set too, but AppKit has no support for that (can be just ignored there).

Also it's still not nice that the MessageBox has no default title on Window, maybe we can add a fall-back using `AssemblyTitleAttribute || AssemblyProductAttribute` (from `Assembly.GetEntryAssembly ()`) rather than use `string.Empty` for the title.

PR #964, opened 2019-08-30
[Mac] Fix Popover backend disposing

sevoku:
If the popover has already been closed, it's delegate
might have been released, so that accessing it would cause
unsupported NSObject resurrection.

Fixes VSTS #803040

### DIFF
--- a/Xwt.XamMac/Xwt.Mac/PopoverBackend.cs
+++ b/Xwt.XamMac/Xwt.Mac/PopoverBackend.cs
@@ -43,6 +43,7 @@ namespace Xwt.Mac
 		public IPopoverEventSink EventSink { get; set; }
 		internal bool EnableCloseEvent { get; private set; }
 		NSPopover popover;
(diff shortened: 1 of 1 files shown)


Xwt.XamMac/Xwt.Mac/PopoverBackend.cs
```diff
@@ -279,14 +280,13 @@ void DestroyPopover ()
 		{
 			if (popover != null) {
 				popover.Close ();
-				var controller = popover.Delegate as FactoryViewController;
```
Therzok: Simpler solution would be to dispose of the delegate before closing?

sevoku: It would be, but see https://github.com/mono/xwt/commit/8904459360f4c359f0304c7da77e1c2de9c441bd, we had issues with that not always being called for some reason.

PR #967, opened 2019-09-01
[Mac] Replace a PInvoke which logs an error at runtime

Therzok:
```
WARNING: The Gestalt selector gestaltSystemVersion is returning 10.9.5 instead of 10.14.5. This is not a bug in Gestalt -- it is a documented limitation. Use NSProcessInfo's operatingSystemVersion property to get correct system version number.
```

Fixes https://github.com/mono/monodevelop/issues/7745

### DIFF
--- a/Xwt.Gtk.Mac/GtkMacDesktopBackend.cs
+++ b/Xwt.Gtk.Mac/GtkMacDesktopBackend.cs
@@ -30,6 +30,7 @@ using System.Drawing;
 using System.IO;
 using GTK = global::Gtk;
 using CoreGraphics;
+using Foundation;
 
 namespace Xwt.Gtk.Mac
 {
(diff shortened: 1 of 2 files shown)


Xwt.Gtk.Mac/GtkMacDesktopBackend.cs
```diff
@@ -30,18 +30,19 @@
 using System.IO;
 using GTK = global::Gtk;
 using CoreGraphics;
+using Foundation;
 
 namespace Xwt.Gtk.Mac
 {
 	public class GtkMacDesktopBackend: GtkDesktopBackend
 	{
 		//this is a BCD value of the form "xxyz", where x = major, y = minor, z = bugfix
```
slluis: This comment doesn't apply anymore.

PR #969, opened 2019-09-20
[Mac] Fix image rendering on Catalina

sevoku:
The Delegate based NSCustomImageRep is broken on Catalina
so that the selector never gets called. We need to use
fixed-size NSCustomImageRep instances for each requested
image size.

Fixes VSTS #983365


sevoku:
Ok the cause for broken icon rendering was a bug in macOS 10.15 Beta 8 and has been fixed with Beta 9, hence closing this.

PR #972, opened 2019-09-30
[Gtk] Fix NRE on LabelBackend's TextIndexer

rodrmoya:
This broke with https://github.com/mono/xwt/commit/2102683398a84ee16094fc4bcae3dac44ba31872#diff-b2fc874d943bfbde146e59ed3a809b6f
which removed the assignment for TextIndexer, so trying to access it,
now that it is never assigned, results in a NRE.

### DIFF
--- a/Xwt.Gtk/Xwt.GtkBackend/LabelBackend.cs
+++ b/Xwt.Gtk/Xwt.GtkBackend/LabelBackend.cs
@@ -39,7 +39,6 @@ namespace Xwt.GtkBackend
 	{
 		Color? textColor;
 		List<LabelLink> links;
-		TextIndexer indexer;
 
 		public LabelBackend ()
 		{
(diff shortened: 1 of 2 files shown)


Xwt.Gtk/Xwt.GtkBackend/LabelBackend.cs
```diff
@@ -168,6 +166,7 @@ public void SetFormattedText (FormattedText text)
 			if (links != null)
 				links.Clear ();
 
+			var indexer = new TextIndexer (text.Text);
```
mrward: Would it make sense to re-use the TextIndexer? The [ApplyFormattedText](https://github.com/mono/xwt/blob/44369a21f3fe3127b325e8c0d61d83da15f1e8e4/Xwt.Gtk/Xwt.GtkBackend/Util.cs#L296) method creates its own TextIndexer. Could we also allow it to be passed to ApplyFormattedText?

rodrmoya: yeah, fixed now to do that

rodrmoya:
@monojenkins backport d16-3

PR #973, opened 2019-09-30
Fix tag filtering when reloading tagged images

sevoku:
A ThemedImage consists of a set of Images with different tags
and the single child images can have several resolutions loaded
from different files.
When switching between toolkits, NativeImageRef.LoadForToolkit
reloads the images for the new toolkit based on NativeImageSource
which holds all data sufficient to reconstruct a ThemedImage.

When performing the actual loading using Image.LoadImage,
several issues occurred:

* NativeImageRef.LoadForToolkit did not pass the tags specified
  for the given source when loading images directly from resources
  or files
* Image.LoadImage always added the base image file even if
  the tag filter did not match. This resulted in creation
  of another nested ThemedImage (with an image group for the specified
  tag and the untagged base image), instead of a simple Image
  with frames for the requested tag.
* ParseImageHints did not detect tags included in the basename
  resulting in missing frames, if the client requested a file
  with tags included.

Fixes VSTS #991543

### DIFF
--- a/Xwt/Xwt.Drawing/Image.cs
+++ b/Xwt/Xwt.Drawing/Image.cs
@@ -171,17 +171,22 @@ namespace Xwt.Drawing
 		/// </remarks>
 		public static Image FromResource (Assembly assembly, string resource)
 		{
+			return FromResource (assembly, resource, null);
+		}
+
+		internal static Image FromResource (Assembly assembly, string resource, ImageTagSet tagFilter)
+		{
 			if (assembly == null)
 				throw new ArgumentNullException ("assembly");
 			if (resource == null)
 				throw new ArgumentNullException ("resource");
-			
+
 			var toolkit = Toolkit.CurrentEngine;
 			if (toolkit == null)
 				throw new ToolkitNotInitializedException ();
 
 			var loader = new ResourceImageLoader (toolkit, assembly);
-			return LoadImage (loader, resource, null);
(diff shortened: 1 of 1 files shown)


Xwt/Xwt.Drawing/Image.cs
```diff
@@ -238,13 +248,14 @@ static bool ParseImageHints (string baseName, string fileName, string ext, out i
 		{
 			scale = 1;
 			tags = ImageTagSet.Empty;
+			var firstDelimiter = fileName.IndexOfAny (new [] { '@', '~' });
```
Therzok: Hoist this outside the method and cache it.

Xwt/Xwt.Drawing/Image.cs
```diff
@@ -238,13 +248,14 @@ static bool ParseImageHints (string baseName, string fileName, string ext, out i
...
 
-			if (!fileName.StartsWith (baseName, StringComparison.Ordinal) || fileName.Length <= baseName.Length + 1 || (fileName [baseName.Length] != '@' && fileName [baseName.Length] != '~'))
+			if (!fileName.StartsWith (baseName, StringComparison.Ordinal) || fileName.Length <= baseName.Length + 1 || firstDelimiter <= 0)
```
Therzok: check firstDelimiter check first?

PR #975, opened 2019-10-01
[Mac] Fix crash due to a missing IntPtr ctor in NSAppearanceCustomizationPopover

sevoku:
Fixes VSTS #935131

### DIFF
--- a/Xwt.XamMac/Xwt.Mac/PopoverBackend.cs
+++ b/Xwt.XamMac/Xwt.Mac/PopoverBackend.cs
@@ -301,6 +301,11 @@ namespace Xwt.Mac
 
 		public class NSAppearanceCustomizationPopover : NSPopover, INSAppearanceCustomization
 		{
+			public NSAppearanceCustomizationPopover ()
+			{ }
+
+			protected NSAppearanceCustomizationPopover (IntPtr handle) : base (handle)
+			{ }
 		}
 	}
 }


Xwt.XamMac/Xwt.Mac/PopoverBackend.cs
```diff
@@ -301,6 +301,11 @@ public static NSRectEdge ToRectEdge (Popover.Position pos)
 
 		public class NSAppearanceCustomizationPopover : NSPopover, INSAppearanceCustomization
 		{
+			public NSAppearanceCustomizationPopover ()
+			{ }
+
+			public NSAppearanceCustomizationPopover (IntPtr handle) : base (handle)
```
Therzok:
```suggestion
			protected NSAppearanceCustomizationPopover (IntPtr handle) : base (handle)
```

Therzok: Why do we need this though? We shouldn't really be adding these constructors...

sevoku: This is the resurrection issue we have in different scenarios. When the NSObject (usually happens to NSWindow or Delegates) gets collected the native object might still be alive. Sometimes Cocoa sends selectors to such zombie objects (might be a timing or sync issue in Cocoa) while they are being released and then XamMac tries to recreate the managed object, which fails due to the missing constructor.

But I wonder now: why can/should that be `protected`?

Therzok: > But I wonder now: why can/should that be protected?

1. You don't want people arbitrarily using the intptr constructor. The intptr constructor doesn't set the managed fields so you end up with an invalid object 99% of the time
2. If we have resurrection issues, it may be worth investigating _why_

If the native object is still alive, sometimes it could be as simple as not explicitly disposing the delegate and letting the toggleref handle that.

This is one of the reasons explicit dispose should be avoided and weakreferences should be used to avoid refcycle.

sevoku: Agreed, we need to review and rewrite most of the code here. I've changed the ctor to protected, so maybe let's just merge this to fix the crash, for now.

PR #977, opened 2019-10-06
Add support for composed images

sevoku:
The new ComposedImage type renders several Images as one.

The only way to create composed images is to use `ImageBuilder` to draw several images to a `VectorImage`, like we do in MonoDevelop: https://github.com/mono/monodevelop/blob/11bd01f8a6260e0ef9c9f5fb6b7fbc7a0e47c191/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide/ImageService.cs#L579-L603

Unfortunately, the ImageBuilder (and so VectorImage) is not cross-tookit capable, because it uses the native [`ContextBackendHandler`](https://github.com/mono/xwt/blob/master/Xwt/Xwt.Backends/ContextBackendHandler.cs#L64) API, by recording native rendering steps (`ImageDescription` for images). This means that a new ImageBuilder needs to be created in the context of each toolkit being used separately.

ComposedImage adds a simple way to workaround this limitation, by storing solely the images and getting the native image backends on demand when drawing, it can be used with any toolkit context.

### DIFF
--- a/Xwt/Xwt.Drawing/ComposedImage.cs
+++ b/Xwt/Xwt.Drawing/ComposedImage.cs
@@ -1,0 +1,67 @@
+﻿//
+// ComposedImage.cs
+//
+// Author:
+//       Vsevolod Kukol <sevoku@microsoft.com>
+//
+// Copyright (c) 2019 Microsoft
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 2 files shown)


Xwt/Xwt.Drawing/ComposedImage.cs
```diff
@@ -0,0 +1,64 @@
...
+
+		public ComposedImage (IEnumerable<Image> images, Size size = default (Size))
+		{
+			this.images = images.ToArray ();
+			Size = size;
+		}
+
+		protected override void OnDraw (Context ctx, Rectangle bounds)
+		{
+			if (images.Length == 0)
```
Therzok: Should we accept at constructor time an empty array? Why have something which does nothing?

Maybe check images for null and empty arrays and throw in that case.

PR #978, opened 2019-10-08
Remove call to label.Destroy()

mcumming:
The MenuItemBackend calls Destroy on a Gtk object it did not create.


garuma:
@monojenkins backport to d16-4

PR #981, opened 2019-10-09
[Bug 992658] Update FileSelector accessibility

GouriKumari:
Fixes: https://devdiv.visualstudio.com/DevDiv/_queries/edit/992658

### DIFF
--- a/Xwt/Xwt/FileSelector.cs
+++ b/Xwt/Xwt/FileSelector.cs
@@ -169,10 +169,11 @@ namespace Xwt
 			var box = new HBox ();
 			box.Accessible.IsAccessible = true;
 			box.Accessible.Role = Accessibility.Role.Group;
-			box.Accessible.Title = Application.TranslationCatalog.GetString("File Selector");
+			box.Accessible.Description = Application.TranslationCatalog.GetString("File Selector");
 
 			entry = new TextEntry ();
 			entry.Accessible.Label = Application.TranslationCatalog.GetString ("Path");
+			entry.Visible = true;
 			entry.Changed += (sender, e) => NotifyFileChange ();
 			box.PackStart (entry, true);
 


GouriKumari:
Screen Shot 2019-10-09 at 6 37 50 PM

@sevoku : https://devdiv.visualstudio.com/DevDiv/_queries/edit/992658 requires a fix in .NET Empty project options. On debugging it looks like the issue is with accessibility of file selector. Could you verify this. Also, can you let me know how to verify these changes
 locally.

rodrmoya:
@GouriKumari just do the change on main/external/xwt, and then build as usual, monodevelop will use the Xwt build from that dir.

sevoku:
@GouriKumari as Rodrigo explained it's a submodule under [main/external/xwt](https://github.com/mono/monodevelop/tree/master/main/external), so you can just pull your changes there. But keep in mind that running `make` will reset it to the upstream commit.

WRT changes: I'm not sure what you are trying to achieve. Theoretically it should be enough to set a Accessibility.Label for those entries in MD code, that's it.

GouriKumari:
@sevoku : I think the issue is here in VSFMac code https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Projects.OptionPanels/AssemblyRunConfigurationEditor.cs#L98 . But the FileSelector code is in xwt, https://github.com/mono/xwt/blob/master/Xwt/Xwt/FileSelector.cs#L165. That is the reason why I tried to change it in xwt.

GouriKumari:
@sevoku : I think I can close this PR. It looks like accessibility is working fine without this fix in 8.3.2. 
https://www.dropbox.com/s/ya4dvn8ro1vihoa/FileSelector%20Save.mov?dl=0 -> This is the screencast when tested from md-addins. 
In 8.3.2 instead of . "Enter File Path" , VO says "Path, you are in an edit text area".

PR #987, opened 2019-10-16
Revert "[Gtk][a11y] Use toggle button in comboboxes for setting a11y."

sevoku:
This fix is not required with latest AtkCocoa

This reverts commit 2abba7b56283ec27cd1632b5e4f10dd1dfc56ae1 (part of #875).


sgmunn:
I haven't been able to verify this yet. Locally for me this breaks VO on my build

sgmunn:
tested this patch with a build from https://github.com/mono/monodevelop/pull/8965 and combo boxes have their accessibility label read out now

PR #990, opened 2019-10-17
[Mac] Focus parent window after selecting a file

sevoku:


Xwt.XamMac/Xwt.Mac/FileDialogBackend.cs
```diff
@@ -64,6 +65,12 @@ protected virtual void OnInitialize (System.Collections.Generic.IEnumerable<File
 		public bool Run (IWindowFrameBackend parent)
 		{
 			var returnValue = Panel.RunModal ();
+			if (parent != null) {
+				var win = parent as NSWindow ?? Runtime.GetNSObject (parent.NativeHandle) as NSWindow;
```
Therzok: can we have an internal extension method on IWindowFrameBackend for this?

sevoku: In fact, we should not even have separate implementations in Xwt.Gtk.Mac

PR #997, opened 2019-10-23
[WPF] Fix Label accessibility

DmitriyKirakosyan:
Wpf's `LabelBackend` uses custom `WpfLabel`, which sets `TextBlock` as
a Label's `Content`, and also redirects the focus to that element. Hence
we should use `WpfLabel.TextBlock` as an a11y target in `AccessibleBackend`.


Xwt.WPF/Xwt.WPFBackend/AccessibleBackend.cs
```diff
@@ -92,7 +92,11 @@ public void EnableEvent (object eventId)
 
 		public void Initialize (IWidgetBackend parentWidget, IAccessibleEventSink eventSink)
 		{
-			Initialize (parentWidget.NativeWidget, eventSink);
+			if (parentWidget.NativeWidget is WpfLabel) {
```
daserge: Looks like the `{}` are not required as per code style around here

PR #1004, opened 2019-10-30
MacEngine: support custom NSApplicationDelegate, improve init

abock:
1. Clean up `NSApplicationInitializer`:
    1. by using `[ThreadStatic]` which is 2-3x faster than `Thread. GetNamedDataSlot` etc
    2. stop using reflection to set `IgnoreMissingAssembliesDuringRegistration` (I think we're far enough into the future to stop paying for this reflection cost on every startup).

2. Support instantiating custom subclasses of XWT's `AppDelegate` so other standard macOS app delegate methods can be handled by the application using XWT. This is necessary, for example, to support OSA scripting.

### DIFF
--- a/Xwt.XamMac/Xwt.Mac/NSApplicationInitializer.cs
+++ b/Xwt.XamMac/Xwt.Mac/NSApplicationInitializer.cs
@@ -24,23 +24,22 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+using System;
+
 using AppKit;
 
 namespace Xwt.Mac
 {
 	static class NSApplicationInitializer
 	{
+		[ThreadStatic]
+		static bool initialized;
+
 		public static void Initialize ()
 		{
-			var ds = System.Threading.Thread.GetNamedDataSlot ("NSApplication.Initialized");
-			if (System.Threading.Thread.GetData (ds) == null) {
-				System.Threading.Thread.SetData (ds, true);
-
-				// IgnoreMissingAssembliesDuringRegistration is only avalilable in Xamarin.Mac 3.4+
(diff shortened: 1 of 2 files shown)


Xwt.XamMac/Xwt.Mac/NSApplicationInitializer.cs
```diff
@@ -24,23 +24,22 @@
...
+using System;
+
 using AppKit;
 
 namespace Xwt.Mac
 {
 	static class NSApplicationInitializer
 	{
+		[ThreadStatic]
+		static bool initialized;
```
Therzok: I don't get why we're using a thread-local storage for this. Shouldn't the app only be initialized on the main thread?

One could potentially call NSApplication.Init on 2 different threads. This seems wrong.

Therzok: Actually calling initialize multiple times makes it throw:
https://github.com/xamarin/xamarin-macios/blob/f1cb31004382bfe2bf87ae0f047534273d06672c/src/AppKit/NSApplication.cs#L52

abock: I did not want to change the semantics of anything. I agree with you though - I do not think you should ever be allowed to call this from multiple threads, but I was going for a perf gain over changing any semantics.

Therzok: Okay, I kind-of understand why it's done like this. Multiple inits from the main thread would not throw.

abock: Yeah, I think the desire was to allow multiple init calls from multiple threads, but allow `NSApplication.Init` itself to enforce multiple-inits _per thread_ by throwing. But again, I don't have the back story as to why it was done like this. I just wanted to modernize the threading semantics in place. /cc @slluis.

PR #1011, opened 2019-11-27
[Mac] Remove ThreadStatic from NSApplicationInitializer.initialized

sevoku:
This allowed NSApplication.Init () to be called several times
from different threads, causing random "System.InvalidOperationException:
Init has already been invoked; it can only be invoked once" exceptions.

Partially reverts 36b5c71a817b6e9c7874f2c0c2de5509c3db2cd0


Therzok:
cc @abock

PR #1012, opened 2019-12-08
Port Xwt to netstandard2.0

Therzok:
This removes System.Xaml reference which was only used to serialize/deserialize xaml.

Port Xwt to ns2.0, and platform assemblies to net472.

### DIFF
--- a/Xwt/Xwt.Design/DesignerSurface.cs
+++ b/Xwt/Xwt.Design/DesignerSurface.cs
@@ -27,7 +27,6 @@
 using System;
 using Xwt.Backends;
 using System.Xml;
-using System.Xaml;
 using Xwt.Drawing;
 
 namespace Xwt.Design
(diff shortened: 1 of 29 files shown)


Therzok:
cc @lambdageek @sevoku

Xwt/Xwt.Design/DesignerSurface.cs
```diff
@@ -61,7 +57,7 @@ public void Load (Widget widget)
 		
 		public void Save (XmlWriter w)
```
Therzok:
```suggestion
		[Obsolete("Xaml loading is not supported", error: true)]
		public void Save (XmlWriter w)
```

Xwt/Xwt.Design/DesignerSurface.cs
```diff
@@ -43,14 +42,11 @@ public DesignerSurface ()
 		IDesignerSurfaceBackend Backend {
 			get { return (IDesignerSurfaceBackend) BackendHost.Backend; }
 		}
-		
+
+		[Obsolete("Xaml loading is not supported", error: true)]
 		public void Load (XmlReader r)
```
Therzok: I'm not sure whether to remove these APIs or not.

Therzok:
Took this for a smoke test, seems to work fine.

PR #1021, opened 2020-01-13
Make ThemedImage style selection less unpredictable

sevoku:
GetImage is optimized for performace and therefore just loops over
the images and selects the first best matching one based on given styles.
The result is not predictable if the image list is not sorted and several alternatives
match different global and explicit styles.
Since we want to prioritize explicit styles first, we need to make sure that we
don't select an image with a given global style but without an explicit match first.
The simplest approach without slowing down rendering by a sophisticated lookup is to
sort images matching global styles last, so that GetImage would always select
an alternative with an explicit rather than a global style.

**[High contrast Mode](https://github.com/mono/monodevelop/pull/9510) Example from MonoDevelop**

We use two problematic/conflicting styles:

* `sel` for icons rendered if a control is in selected state
* `contrast` icon alternatives used when a high contrast shell theme is active

The `sel` mode is always explicitely requested if the control state changes, while the `contrast` style is global.

Given we have these two icons for the different modes:

* `icon~sel.png`
* `icon~contrast.png`

It is not well defined which one would be rendered, if both styles are set. Currently this depends only on the order in which the icon alternatives have been loaded from disk. With this change we'll always select the explicit and not the global style, which is `sel` in this case.

without the fix:
grafik
with the fix:
grafik

### DIFF
--- a/Xwt/Xwt.Drawing/ThemedImage.cs
+++ b/Xwt/Xwt.Drawing/ThemedImage.cs
@@ -31,7 +31,9 @@ namespace Xwt.Drawing
 {
 	public class ThemedImage: DrawingImage
 	{
-		List<Tuple<Image, string []>> images;
+		readonly List<Tuple<Image, string []>> images;
+
+		int globalStylesVersion = -1;
 
 		public ThemedImage (List<Tuple<Image, string []>> images, Size size = default (Size))
 		{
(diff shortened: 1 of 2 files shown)


Xwt/Xwt.Drawing/ThemedImage.cs
```diff
@@ -47,7 +81,7 @@ public ThemedImage (List<Tuple<Image, string []>> images, Size size = default (S
 
 		protected override void OnDraw (Context ctx, Rectangle bounds)
 		{
-			var best = GetImage (ctx.Styles);
+			var best = GetImage (ctx.Styles.OrderBy (s => Context.HasGlobalStyle (s)));
```
Therzok: This is too expensive to do on draw.

sevoku: Ups, that was unintended.

slluis:
I'd like to suggest a slightly different approach to avoid the event subscription. Have a global counter that is increased when there is a change in the global styles. Store a copy of the current value of the counter in the themed image when it is created. Before rendering an image, check if the counter changed since last draw, and in that case resort the images before rendering.

PR #1030, opened 2020-02-13
[WPF] fixes a multiple call to OnSelectedItemsChanged in TreeView

paymicro:
When you select an item in a TreeView `OnSelectedItemsChanged` is called multiple times.


paymicro:
@sevoku take a look please.

PR #1031, opened 2020-02-27
[GTK] Unsubscribe Popover from Hidden event on hiding

daserge:
In order to test replace
```
		void PopoverHidden (object sender, EventArgs e)
		{
			sink.OnClosed ();
			popover.Hidden -= PopoverHidden;
		}
```
with
```
		void PopoverHidden (object sender, EventArgs e)
		{
			Console.WriteLine ("Backend  Popover_Hidden event");
			sink.OnClosed ();
			//popover.Hidden -= PopoverHidden;
		}
```
then open Sample -> Popover page and open and close a button to trigger Popover, observe how console accumulates the messages with each Show

### DIFF
--- a/Xwt.Gtk/Xwt.GtkBackend/PopoverBackend.cs
+++ b/Xwt.Gtk/Xwt.GtkBackend/PopoverBackend.cs
@@ -341,6 +341,8 @@ namespace Xwt.GtkBackend
 			if (popover.TransientFor != null)
 				popover.TransientFor.FocusInEvent -= HandleParentFocusInEvent;
 
+			popover.Hidden -= PopoverHidden;
+
 			popover.Destroy ();
 			popover.Dispose ();
 		}
--- a/Xwt.Gtk/Xwt.GtkBackend/PopoverBackend.cs
+++ b/Xwt.Gtk/Xwt.GtkBackend/PopoverBackend.cs
@@ -269,7 +269,7 @@ namespace Xwt.GtkBackend
 				popover.TransientFor.FocusInEvent += HandleParentFocusInEvent;
 			}
 
-			popover.Hidden += (o, args) => sink.OnClosed ();
+			popover.Hidden += PopoverHidden;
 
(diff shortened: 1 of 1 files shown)


Xwt.Gtk/Xwt.GtkBackend/PopoverBackend.cs
```diff
@@ -269,7 +269,7 @@ public void Show (Xwt.Popover.Position orientation, Xwt.Widget reference, Xwt.Re
 				popover.TransientFor.FocusInEvent += HandleParentFocusInEvent;
 			}
 
-			popover.Hidden += (o, args) => sink.OnClosed ();
+			popover.Hidden += PopoverHidden;
```
netonjm: unsuscribe on dispose

sevoku: not sure, won't popover.Destroy () raise the event and unsbscribe? @daserge could you verify?

netonjm: ahh ! yeah! sorry I thought it was the Cocoa backend 👍

daserge: Maybe I don't understand the suggestion but this still has the issue:
Screen Shot 2020-02-27 at 9 09 12 PM
and if I change the sample code to `using (popover = new Popover ()) { ... popover.Show (...) }` it doesn't show up

sevoku: not `using`, the backend will be disposed when you close the window, or add another button to dispose it and see if it calls PopoverHidden there

daserge: OK, so if I create a temporary Popover in button click handler it does call Dispose on closing.
The question is then - is the sample a legitimate usage?
It stores a variable in the field so that another Show does another subscription to Hidden.

PR #1037, opened 2020-05-13
Fixed canvas double drawing and made `BackgroundColor` virtual

mykyta-bondarenko:


Xwt.XamMac/Xwt.Mac/CanvasBackend.cs
```diff
@@ -52,6 +52,8 @@ protected override void OnSizeToFit ()
 			Widget.SetFrameSize (new CGSize ((nfloat)s.Width, (nfloat)s.Height)); 
 		}
 
+		public override Xwt.Drawing.Color BackgroundColor { get; set; }
```
sevoku, line 55:
```suggestion
		// Skip Layer initialization in base to avoid rendering issues. CanvasView.DrawRect will draw the background.
		public override Xwt.Drawing.Color BackgroundColor { get; set; }
```

mykyta-bondarenko:
@sevoku we can keep it as it's but use `new` word to make a workaround in vsm code

sevoku:
Reopening, we still need to fix this.

PR #1047, opened 2021-02-16
Fix incorrect alternative images being used

mrward:
With embedded resources:

gtc-solution.png
gtc-solution@2x.png
gtc-solution~contrast.png

gtc-solution-filter.png
gtc-solution-filter@2x.png
gtc-solution-filter~contrast.png

The ParseImageHints method would find an image match when the
baseName was gtc-solution and the fileName was gtc-solution@2x.png
or gtc-solution~contrast.png. The problem was the original check
used fileName.StartsWith(baseName) which allowed this mismatch.
Changing the check to compare the fileName before the first
delimiter found with the baseName avoids an incorrect image being
used.

### DIFF
--- a/Xwt/Xwt.Drawing/Image.cs
+++ b/Xwt/Xwt.Drawing/Image.cs
@@ -252,7 +252,7 @@ namespace Xwt.Drawing
 			tags = ImageTagSet.Empty;
 			var firstDelimiter = fileName.IndexOfAny (tagDelimiters);
 
-			if (firstDelimiter <= 0 || fileName.Length <= baseName.Length + 1 || !fileName.StartsWith (baseName, StringComparison.Ordinal))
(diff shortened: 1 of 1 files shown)


sevoku (approved):

Xwt/Xwt.Drawing/Image.cs
```diff
@@ -252,7 +252,7 @@ static bool ParseImageHints (string baseName, string fileName, string ext, out i
 			tags = ImageTagSet.Empty;
 			var firstDelimiter = fileName.IndexOfAny (tagDelimiters);
 
-			if (firstDelimiter <= 0 || fileName.Length <= baseName.Length + 1 || !fileName.StartsWith (baseName, StringComparison.Ordinal))
+			if (firstDelimiter <= 0 || fileName.Length <= baseName.Length + 1 || !fileName.Substring(0, firstDelimiter).Equals(baseName, StringComparison.Ordinal))
```
Therzok, line 255: This introduces an allocation here which is probably costly, especially considering how often this is called.

We probably want some other way to validate this.

PR #1051, opened 2021-03-22
Allow application to specify whether to initialize parent toolkit or not

iainx:
If Xwt.Mac is initialized inside a Cocoa application then
NSApplication.Initialize cannot be called twice.

### DIFF
--- a/Xwt.XamMac/Xwt.Mac/MacEngine.cs
+++ b/Xwt.XamMac/Xwt.Mac/MacEngine.cs
@@ -47,20 +47,29 @@ namespace Xwt.Mac
 		
 		public override void InitializeApplication ()
 		{
-			NSApplicationInitializer.Initialize ();
+			if (InitializeToolkit)
+			{
+				NSApplicationInitializer.Initialize();
 
-			//Hijack ();
-			if (pool != null)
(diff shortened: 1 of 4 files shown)


Xwt.XamMac/Xwt.Mac/MacEngine.cs
```diff
@@ -47,20 +47,29 @@ public class MacEngine: Xwt.Backends.ToolkitEngineBackend
...
+				// the application without a bundle
+				var info = NSBundle.MainBundle.InfoDictionary;
+				if (info.ValueForKey((NSString)"NSPrincipalClass") == null)
+					info.SetValueForKey((NSString)"NSApplication", (NSString)"NSPrincipalClass");
+			}
+			else
+			{
+				// Although the AppDelegate does not need to be hijacked, keep this allocated
+				// as there are some non-NSApplicationDelegate methods still used by Xwt
+				appDelegate = new AppDelegate(IsGuest);
```
Therzok, line 71: Wondering why we wouldn't want AppDelegateFactory to be called int his case.

Therzok (approved):

sevoku (approved):

PR #1052, opened 2021-04-22
[Mac] Manually register assemblies for Xamarin.Mac

Therzok:
This patch allows for two main improvements:

* The entry assembly and its dependencies are no longer force-loaded
* We register assemblies that are not referenced by the entry assembly
  and are loaded later on by the app

### DIFF
--- a/Xwt.XamMac/Xwt.Mac/NSApplicationInitializer.cs
+++ b/Xwt.XamMac/Xwt.Mac/NSApplicationInitializer.cs
@@ -24,7 +24,11 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+using System;
+
 using AppKit;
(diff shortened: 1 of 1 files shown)


Xwt.XamMac/Xwt.Mac/NSApplicationInitializer.cs
```diff
@@ -36,14 +40,34 @@ public static void Initialize ()
...
+				Runtime.AssemblyRegistration += Runtime_AssemblyRegistration;
+
 				NSApplication.Init ();
+
+				// Register a callback when an assembly is loaded so it's registered in the xammac registrar.
+				AppDomain.CurrentDomain.AssemblyLoad += CurrentDomain_AssemblyLoad;
+
+				// Manually register all the currently loaded assemblies.
+				foreach (var assembly in AppDomain.CurrentDomain.GetAssemblies())
+                {
```
KirillOsenkov, line 55: whitespace but don't block on this

PR #1054, opened 2021-05-06
Make Registrar code thread-safe on our end

Therzok:


slluis (approved):

Therzok:
Ideally, Xamarin.Mac should have some thread safety, but since we disable the startup loading and hook our own, this is good enough.

PR #1055, opened 2021-05-19
Allow specifying an exit code when exiting the application

slluis:
NSApplication.Main never returns. The only way to return an exit code is to call exit(code).

### DIFF
--- a/Xwt.XamMac/Xwt.Mac/MacEngine.cs
+++ b/Xwt.XamMac/Xwt.Mac/MacEngine.cs
@@ -155,6 +155,14 @@ namespace Xwt.Mac
 			NSApplication.SharedApplication.Terminate(appDelegate);
 		}
 
+		public override void ExitApplication(int exitCode)
+		{
+			if (exitCode == 0)
+				NSApplication.SharedApplication.Terminate(appDelegate);
+			else
+				Environment.Exit(exitCode);
+		}
+
 		static Selector hijackedSel = new Selector ("hijacked_loadNibNamed:owner:");
 		static Selector originalSel = new Selector ("loadNibNamed:owner:");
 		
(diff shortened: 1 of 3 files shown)


Xwt.XamMac/Xwt.Mac/MacEngine.cs
```diff
@@ -155,6 +155,14 @@ public override void ExitApplication ()
 			NSApplication.SharedApplication.Terminate(appDelegate);
 		}
 
+		public override void ExitApplication(int exitCode)
+		{
+			if (exitCode == 0)
+				NSApplication.SharedApplication.Terminate(appDelegate);
+			else
+				Environment.Exit(exitCode);
```
Therzok, line 163: This is not 100% correct. Using Environment.Exit will almost always be a slow quit.

Therzok: Just wanted to mention this, since last time I tried it, it took 10 seconds to quit

slluis: Is there any other way of reporting the exit code?

Therzok: https://developer.apple.com/documentation/appkit/nsapplication/1428417-terminate?language=objc

Not that I can find.

Also, calling Environment.Exit will prevent normal ShouldTerminate/WillTerminate handlers from running, afaik.

Given app bundles are supposed to be run by launch services, not directly launched, an exit code does not make sense, as it's not observable, I guess?

slluis: We need it since we need a result when running unit tests through vstool

PR #1056, opened 2021-06-02
[Xwt] Exit user code when initialising as guest

iainx:
When used as a guest toolkit, there needs to be a call to ExitUserCode or non
of the dispatch code works


slluis:
I'm curious, what's this for?

slluis (approved):

PR #1060, opened 2021-06-30
[Mac] Fix TableView layout and rendering on BigSur

sevoku:
Big Sur introduced a new NSTableViewStyle and additional weird behavior
breaking the layout and CanvasCell rendering.

### DIFF
--- a/Xwt.XamMac/Xwt.Mac/TableViewBackend.cs
+++ b/Xwt.XamMac/Xwt.Mac/TableViewBackend.cs
@@ -446,6 +446,19 @@ namespace Xwt.Mac
 				// may be drawn already and it will not be redrawn even
 				// if Selection has been changed.
 				NeedsDisplay = true;
+				// Since Big Sur NSTableView doesn't refresh its subviews on selection change
+				if (MacSystemInformation.OsVersion >= MacSystemInformation.BigSur)
+				{
+					// We have always a CompositeCell with Subviews for real cell views
+					foreach (var sub in Subviews)
+					{
+						sub.NeedsDisplay = true;
+						foreach (var subsub in sub.Subviews)
+						{
+							subsub.NeedsDisplay = true;
+						}
+					}
+				}
 			}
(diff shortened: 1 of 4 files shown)


Xwt.XamMac/Xwt.Mac/TableViewBackend.cs
```diff
@@ -446,6 +446,19 @@ class TableRowView : NSTableRowView
 				// may be drawn already and it will not be redrawn even
 				// if Selection has been changed.
 				NeedsDisplay = true;
+				// Since Big Sur NSTableView doesn't refresh its subviews on selection change
+				if (MacSystemInformation.OsVersion >= MacSystemInformation.BigSur)
```
Therzok, line 450: This is very very weird, expensive and horrible.

Therzok: Isn't it because of the Mojave change with layer backed views?
https://developer.apple.com/documentation/macos-release-notes/appkit-release-notes-for-macos-10_14

Therzok: > Views that implicitly depend on being redrawn when an ancestor, descendant, or intersecting sibling is redrawn may not be redrawn. As before, if a view needs to be redrawn, set its needsDisplay property to true.

Specifically.

sevoku: No, this can be only observed on BigSur, also no layers involved here. I have no idea how to make this in a better way. Ideas?

Therzok: Only thing I can find: https://developer.apple.com/forums/thread/659488

sevoku: Filed an internal bug for bad code. Per discussion on Teams let's merge this.

PR #1061, opened 2021-07-08
[Mac] Fix TableView selection rendering when building with Xcode 12

sevoku:
This removes the Big Sur check from #1060, since the trigger for this issue is not Big Sur but building with Xcode 12.


Therzok (approved):
So it was an SDK issue :D

PR #1075, opened 2021-11-24
Adds fullpaths to dylibs in libglib and pango to fix monterey linking

netonjm:


HinTak:
I don't think the path in dll.config has got anything to do with monterey linking? I think it is better to let the linker decide (to allow embedded and relocated usage)

netonjm:
> I don't think the path in dll.config has got anything to do with Monterey linking? I think it is better to let the linker decide (to allow embedded and relocated usage)

Just to add some context, after the Monterey update, the linking is failing to get correct paths in some users with Homebrew installed and packages with libcairo and libglib dependencies. The problem is that the linking is incorrectly deciding the directories to look for, instead of using the Mono folder as the first option, it uses the libraries within Homebrew, and this causes an NRE at startup. That is a workaround to avoid this scenario when starting, but I agree it is something that should be solved in a general way in MONO.

Xwt.Gtk/Xwt.Gtk.dll.config
```diff
@@ -10,13 +10,13 @@
...
-	<dllmap os="osx" dll="libgdk-win32-2.0-0.dll" target="libgdk-quartz-2.0.0.dylib"/>
-	<dllmap os="osx" dll="libpango-1.0-0.dll" target="libpango-1.0.0.dylib"/>
-	<dllmap os="osx" dll="libpangocairo-1.0-0.dll" target="libpangocairo-1.0.0.dylib"/>
+	<dllmap os="osx" dll="libglib-2.0-0.dll" target="/Library/Frameworks/Mono.framework/Versions/Current/lib/libglib-2.0.0.dylib"/>
+	<dllmap os="osx" dll="libgobject-2.0-0.dll" target="/Library/Frameworks/Mono.framework/Versions/Current/lib/libgobject-2.0.0.dylib"/>
+	<dllmap os="osx" dll="libatk-1.0-0.dll" target="/Library/Frameworks/Mono.framework/Versions/Current/lib/libatk-1.0.0.dylib"/>
+	<dllmap os="osx" dll="libgtk-win32-2.0-0.dll" target="/Library/Frameworks/Mono.framework/Versions/Current/lib/libgtk-quartz-2.0.0.dylib"/>
+	<dllmap os="osx" dll="libgdk-win32-2.0-0.dll" target="/Library/Frameworks/Mono.framework/Versions/Current/lib/libgdk-quartz-2.0.0.dylib"/>
+	<dllmap os="osx" dll="libpango-1.0-0.dll" target="/Library/Frameworks/Mono.framework/Versions/Current/lib/libpango-1.0.0.dylib"/>
+	<dllmap os="osx" dll="libpangocairo-1.0-0.dll" target="/Library/Frameworks/Mono.framework/Versions/Current/lib/libpangocairo-1.0.0.dylib"/>
```
akoeplinger, line 19: you can use the special value `$mono_libdir` instead to avoid hardcoding the path, the mono runtime will resolve the path to the `lib` directory at runtime then.

netonjm: Good idea @akoeplinger , thanks for the tip!

Therzok:
Funny thing is that it was a guess :D

Therzok (approved):

sevoku:
I have merged the changes to d16-10 branch, we're using now for the next VSMac 8.10 release. I don't think this needs to be in main and I agree with @HinTak about linker. Converting this to a draft for now.

PR #1080, opened 2022-02-03
Xwt.XamMac: Only reference XM when not a net6 build

sandyarmstrong:


sandyarmstrong:
I see now this is pretty similar to https://github.com/mono/xwt/tree/dev/therzok%2Fnet6-macos-cleanup. No preference about the approach.

Therzok (approved):

PR #1081, opened 2022-02-04
Delegate fix

leculver:
This fixes one of VS for Mac's top crashing buckets **System.ExecutionEngineException_MonoDevelop.PackageManagement.dll!MonoDevelop.PackageManagement.ManagePackagesDialog::SelectProjects**.

**PLEASE NOTE:**  I have not been able to build or test this change, as I don't work in this space and couldn't get the branch to compile.  This is a proposed change to illustrate the correct fix, but I was not able to get this repo to successfully build locally.  This change needs to be built and tested by a VS for Mac/XWT/Xamarin/etc dev before merging.

VS for Mac is encountering a fatal error of: A callback was made on a garbage collected delegate of type 'Xwt.XamMac!Xwt.Mac.ViewBackend+DelegateIntPtrIntPtrBool::Invoke'.

The underlying root cause for this is that the delegates that are created and passed off to native code are not being kept alive.  The GC has no way of knowing that delegates passed to native code need to be kept alive so a root must be established in some way.

Since these are all static methods, we can create a static variable to hold these delegates for the lifetime of the application.

This looks like a regression introduced by commit ba3047716627220e542d06f4e914f796ef46214c.

### DIFF
--- a/Xwt.XamMac/Xwt.Mac/ViewBackend.cs
+++ b/Xwt.XamMac/Xwt.Mac/ViewBackend.cs
@@ -535,6 +535,13 @@ namespace Xwt.Mac
 		static Selector becomeFirstResponderSel = new Selector ("becomeFirstResponder");
 		static Selector resignFirstResponderSel = new Selector ("resignFirstResponder");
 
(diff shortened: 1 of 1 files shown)


leculver:
@slluis Can you please take a look at this pull request?  This relates to a change you made.

@kdubau This is the proposed fix for the crash bucket we discussed today.

Therzok (approved):
LGTM! Thanks!

sandyarmstrong (approved):

PR #1087, opened 2022-02-23
Cocoa passes ctrl+i as a tab key, so check that Characters is not whitespace

iainx:


sandyarmstrong (approved):
Weird.

iainx:
Just to confirm that it's not something we're doing, I've written a test objc cocoa app and it gets tab for ctrl-i as well

Xwt.XamMac/Xwt.Mac/KeyboardUtil.cs
```diff
@@ -128,7 +128,7 @@ static Key GetXwtKey (NSEvent keyEvent, out NSEventModifierMask modMask)
 			// If you press the cmd+м on the keyboard, keyEvent.Characters contains a 'v' and keyEvent.CharactersIgnoringModifiers
 			// contains the м
 			// This way it can map cmd+м to paste like on a roman keyboard.
-			var characters = keyEvent.Characters;
+			var characters = string.IsNullOrWhiteSpace(keyEvent.Characters) ? keyEvent.CharactersIgnoringModifiers : keyEvent.Characters;
```
Therzok, line 131: We're marshalling the Characters string twice now.

PR #1089, opened 2022-02-24
Add a hardware key field to get the physically pressed key

iainx:
Parse the hardware key so we can use it for keyboard shortcuts.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1418403

### DIFF
--- a/Xwt.XamMac/Xwt.Mac/KeyboardUtil.cs
+++ b/Xwt.XamMac/Xwt.Mac/KeyboardUtil.cs
@@ -34,9 +34,9 @@ namespace Xwt.Mac
 		public static KeyEventArgs ToXwtKeyEventArgs (this NSEvent keyEvent)
 		{
 			NSEventModifierMask mask;
-			Key key = GetXwtKey(keyEvent, out mask, out var rawKey);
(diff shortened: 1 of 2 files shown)


sandyarmstrong (approved):
Approving but I'm a bit concerned, will this hold up with various international keyboard layouts?

iainx:
So, this handles everything for what I have, switching the keyboard layout for Cyrillic (for example) produces a myriad other symbols that aren't handled. so in English-GB layout alt-s is ß and that's handled and is mapped to alt+s. In Cyrillac alt-s is ы and that's not handled, so it's not fixed for that, but short of working out all the mappings for all keyboards, I'm not sure what further steps we can take continuing along this path of using Characters/CharactersWithoutModifiers.

Therzok (approved):

Therzok (approved):

Therzok (approved):

sandyarmstrong (approved):

iainx:
So instead of adding a RawKey value, I've made the Key be the US value and if we need to get the unicode value, we should check characters, because ultimately, that's what was happening with the multiple switch statements anyway

sandyarmstrong (approved):

PR #1090, opened 2022-03-01
Fix Cocoa link labels to properly launch the browser

BretJohnson:
Update XamMac link labels to launch the browser directly,
in the standard way we do it elsewhere in VSMac.

Fixes  https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1484767

This behavior worked before with the Gtk backend, but
I don't think it ever worked properly with the Cocoa

### DIFF
--- a/Xwt.XamMac/Xwt.Mac/LinkLabelBackend.cs
+++ b/Xwt.XamMac/Xwt.Mac/LinkLabelBackend.cs
@@ -102,7 +102,7 @@ namespace Xwt.Mac
 		void HandleClicked (object sender, EventArgs e)
 		{
 			ApplicationContext.InvokeUserCode (() => {
-				EventSink.OnNavigateToUrl (uri);
+				NSWorkspace.SharedWorkspace.OpenUrl (uri);
 			});
 		}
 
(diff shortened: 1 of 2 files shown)


Xwt.XamMac/Xwt.Mac/LinkLabelBackend.cs
```diff
@@ -102,7 +102,7 @@ public override void DisableEvent (object eventId)
 		void HandleClicked (object sender, EventArgs e)
 		{
 			ApplicationContext.InvokeUserCode (() => {
-				EventSink.OnNavigateToUrl (uri);
+				NSWorkspace.SharedWorkspace.OpenUrl (uri);
```
mrward, line 105: Not sure this is correct. This means the OnNavigateToUrl event is no longer used. The LinkLabel can subscribe to this and take custom action.

Does this not get called?

https://github.com/mono/xwt/blob/d89687fd4d3885b42f3aeca2e82e0c54dc4fb7d9/Xwt/Xwt/LinkLabel.cs#L123

Therzok: Yeah, this breaks all the custom link implementation.

BretJohnson: Thx. I fixed it to override the desktop backend OpenUrl instead. Please review the latest.

mrward (approved):

Therzok (approved):

sandyarmstrong (approved):

iainx (approved):

PR #1091, opened 2022-03-01
Fix drag/drop support

BretJohnson:
Fix so that drag/drop support works properly with the Cocoa backend.
This PR includes these fixes:

1. DraggingUpdate passed the wrong coordinates. Now do the proper conversion from view coordinates to window coordinates, as expected by the Xwt client.
2. Get the INSDraggingInfo object in the proper way, with `Runtime.GetINativeObject<INSDraggingInfo> (dragInfo, owns: false)`, same as we do in other code now.
3. Use the newer NSView.BeginDraggingSession API instead of the deprecated NSView.DragImage. This is needed so that we can provide a DraggingSource object in order to be notified when the drag/drop is over (no matter if successful or not) and call the Finished callback, as expected by Xwt client code.
4. Populate the pasteboard item used for the drag properly, supporting text drag/drop and internal formats (normally a serialized object where the Id is the assembly qualified type name).

To my knowledge the Android Designer is the only client that still uses this drag/drop code. All of these fixes were to make Android Designer drag/drop work properly.

### DIFF
--- a/Xwt.XamMac/Xwt.Mac/ViewBackend.cs
+++ b/Xwt.XamMac/Xwt.Mac/ViewBackend.cs
@@ -620,7 +620,7 @@ namespace Xwt.Mac
 
 			var di = Runtime.GetINativeObject<INSDraggingInfo> (dragInfo, owns: false);
 			var types = di.DraggingPasteboard.Types.Select (ToXwtDragType).ToArray ();
-			var pos = new Point (di.DraggingLocation.X, di.DraggingLocation.Y);
+			var pos = backend.Widget.ConvertPointFromView (di.DraggingLocation, null).ToXwtPoint ();
 			
 			if ((backend.currentEvents & WidgetEvent.DragOverCheck) != 0) {
 				var args = new DragOverCheckEventArgs (pos, types, ConvertAction (di.DraggingSourceOperationMask));
--- a/Xwt.XamMac/Xwt.Mac/ViewBackend.cs
+++ b/Xwt.XamMac/Xwt.Mac/ViewBackend.cs
@@ -724,19 +724,21 @@ namespace Xwt.Mac
 		void InitPasteboard (NSPasteboard pb, TransferDataSource data)
 		{
 			pb.ClearContents ();
+
 			foreach (var t in data.DataTypes) {
 				// Support dragging text internally and externally
 				if (t == TransferDataType.Text) {
 					pb.AddTypes(new string[] { NSPasteboard.NSStringType }, null);
 					pb.SetStringForType((string)data.GetValue(t), NSPasteboard.NSStringType);
 				}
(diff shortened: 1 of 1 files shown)


Xwt.XamMac/Xwt.Mac/ViewBackend.cs
```diff
@@ -617,10 +617,10 @@ static NSDragOperation DraggingUpdated (IntPtr sender, IntPtr sel, IntPtr dragIn
 			if (ob == null)
 				return NSDragOperation.None;
 			var backend = ob.Backend;
-			
-			INSDraggingInfo di = (INSDraggingInfo) Runtime.GetNSObject (dragInfo);
+
+			var di = Runtime.GetINativeObject<INSDraggingInfo> (dragInfo, owns: false);
```
Therzok, line 621: These changes LGTM!

Therzok (approved):

Xwt.XamMac/Xwt.Mac/ViewBackend.cs
```diff
@@ -883,5 +909,60 @@ public void UpdateChildPlacement ()
 			child.Frame = new CGRect ((nfloat)cx, (nfloat)cy, (nfloat)cwidth, (nfloat)cheight);
 		}
 	}
+
+	class DraggingSource : NSDraggingSource
```
Therzok, line 913:
```suggestion
	class DraggingSource : INSDraggingSource
```

BretJohnson: Fixed (now it's `class DraggingSource : NSObject, INSDraggingSource`.  I believe that's correct (and it seems to work) but feel free to confirm.

BretJohnson:
Removing draft, as this should finally be ready to merge. @netonjm , @sandyarmstrong  - can you please also review.

Therzok (approved):

sandyarmstrong (approved):
Great work, thanks for this!

sandyarmstrong:
I'll let @netonjm coordinate merging/bumping.

BretJohnson:
@sevoku - This should be ready to merge, as soon as you approve.
Once merged, I'll include it in the Xwt bump here: https://github.com/xamarin/vsmac/pull/6485.

BretJohnson:
@netonjm - Did you want @sevoku to look at anything in particular here? This has gotten a few other reviews already & I believe we're good.
In any case, it would great to get this merged as soon as we can, as not having it is something of a blocker for us. Thanks.

netonjm:
> @netonjm - Did you want @sevoku to look at anything in particular here? This has gotten a few other reviews already & I believe we're good.
> In any case, it would great to get this merged as soon as we can, as not having it is something of a blocker for us. Thanks.

I added to add a Final review, but from my side is ready to go

PR #1096, opened 2022-03-16
[AlertDialogBackend] Ensures parent is set using the active window if there are not any window specified 

netonjm:
In cocoa backend when opening an modal AlertBox we need specify a parent and center to it for a correct positioning and avoid odd behaviours.

### DIFF
--- a/Xwt.XamMac/Xwt.Mac/AlertDialogBackend.cs
+++ b/Xwt.XamMac/Xwt.Mac/AlertDialogBackend.cs
@@ -112,6 +112,19 @@ namespace Xwt.Mac
 			Window.ReleasedWhenClosed = true;
 			if (win != null)
 				return sortedButtons [(int)this.RunSheetModal (win) - 1000];
+
+			if (win == null)
+			{
+				//a modal dialog needs parent window so we try take the current key
+				win = NSApplication.SharedApplication.ModalWindow ?? NSApplication.SharedApplication.KeyWindow;
+			}
+			
+			if (win != null)
+			{
+				win.AddChildWindow(this.Window, NSWindowOrderingMode.Above);
+			}
+
+			Util.CenterWindow(this.Window, win);
 			return sortedButtons [(int)this.RunModal () - 1000];
 		}
 


Xwt/Xwt/MessageDialog.cs
```diff
@@ -305,7 +305,8 @@ static Command GenericAlert (WindowFrame parent, MessageDescription message)
 			using (backend) {
 				Command res = null;
 				Toolkit.CurrentEngine.InvokePlatformCode (delegate {
-					res = backend.Run (parent ?? RootWindow, message);
+					WindowFrame parentWindow = parent ?? RootWindow ?? Toolkit.CurrentEngine.GetActiveWindow() as WindowFrame;
+					res = backend.Run (parentWindow, message);
```
Therzok, line 309: It sounds like something the toolkit could implement instead of having a GetActiveWindow() API.

Therzok: `IAlertDialogBackend` should be key/focused window aware

Therzok: In example, I think it's better to just implement it here:
https://github.com/mono/xwt/blob/83ebb134aa9457313bac90c786fc8201dfa7e1b5/Xwt.XamMac/Xwt.Mac/AlertDialogBackend.cs#L111

netonjm: Yeah, I was thinking on this yesterday, in fact in other toolkits this probably doesn't makes sense the parenting works in a different way. I moved the logic to

Therzok (approved):

Xwt.XamMac/Xwt.Mac/AlertDialogBackend.cs
```diff
@@ -112,9 +112,41 @@ public Command Run (WindowFrame transientFor, MessageDescription message)
...
 
+		static void CenterWindow(NSWindow nsChild, NSWindow nsParent)
+		{
+			if (nsParent != null)
+			{
+				if (!nsParent.IsVisible)
+				{
+					nsChild.Center();
+					return;
+				}
```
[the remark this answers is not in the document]
netonjm: fixed

sevoku (approved):

Therzok (approved):

jstedfast (approved):

PR #1097, opened 2022-03-17
[Xam.Mac] Fixes parenting in DialogBackend

netonjm:
This PR fixes set the correct parent when a dialog is shown in modal in the CocoaBackend and it also ensures it's centered to it.

### DIFF
--- a/Xwt.XamMac/Xwt.Mac/DialogBackend.cs
+++ b/Xwt.XamMac/Xwt.Mac/DialogBackend.cs
@@ -169,6 +169,13 @@ namespace Xwt.Mac
 		{
 			Visible = true;
 			modalSessionRunning = true;
+
+			NSWindow nsParent = parent.Window as NSWindow;
+			if (nsParent != null && nsParent.IsVisible)
+			{
+				nsParent.AddChildWindow(this, NSWindowOrderingMode.Above);
+			}
+			Util.CenterWindow(this, nsParent);
 			NSApplication.SharedApplication.RunModalForWindow (this);
 		}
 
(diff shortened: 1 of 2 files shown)


Xwt.XamMac/Xwt.Mac/DialogBackend.cs
```diff
@@ -169,6 +169,13 @@ public void RunLoop (IWindowFrameBackend parent)
 		{
 			Visible = true;
 			modalSessionRunning = true;
+
+			if (parent.Window is NSWindow nsParent)
```
Therzok, line 173: Shoudl we add it as a child window if the parent is not visible?

Xwt.XamMac/Xwt.Mac/DialogBackend.cs
```diff
@@ -169,6 +169,13 @@ public void RunLoop (IWindowFrameBackend parent)
...
+            {
+				nsParent.AddChildWindow(this, NSWindowOrderingMode.Above);
+				Util.CenterWindow(this, nsParent);
```
Therzok, line 176: Don't we also want to center when not parented?

netonjm: fixed

Xwt.XamMac/Xwt.Mac/DialogBackend.cs
netonjm: fixed

Xwt.XamMac/Xwt.Mac/DialogBackend.cs
```diff
@@ -169,6 +169,13 @@ public void RunLoop (IWindowFrameBackend parent)
 		{
 			Visible = true;
 			modalSessionRunning = true;
+
+			NSWindow nsParent = parent.Window as NSWindow;
+			if (nsParent != null && nsParent.IsVisible)
+            {
```
mrward, line 175: Minor: Indentation

mrward (approved):

Xwt.XamMac/Xwt.Mac/DialogBackend.cs
netonjm: fixed

mkrueger (approved):

PR #1098, opened 2022-03-25
Fix password autofill and wrapping

kunyli:
This PR fixes the following issues with the VSM add apple developer account dialog:
 -autofill password does not enable the sign in button because the password from Apple Keychain is not being detected by the change handler
-long passwords wrap onto multiple lines


jstedfast (approved):

Therzok:
Not a problem in this PR because Xwt is full of leaks ( `Activated` was already used), but EventHandler usage in Xamarin.Mac generally leaks, no matter what you do.

Therzok:
Only reliable pattern I've found while investigating this was  subscribe in NSViewController.ViewDidAppear and unsubscribe in ViewDidDisappear

PR #1100, opened 2022-04-06
[WindowFrameBackend] Only add parenting when dialog is preparing it be shown

netonjm:
This PR fixes some issues in WindowFrameBackend:
* NRE whe transientFor is null (we added to cover this scenario)
* SetTransientFor tries to set the window passed as a parameter in the funcion and in case is not the current one it sets the value.
* Simplifies InternalShow() now takes the parent window and if is Visible and not contains the Window it adds as a children.

SoN1wTqqfr

### DIFF
--- a/Xwt.XamMac/Xwt.Mac/WindowFrameBackend.cs
+++ b/Xwt.XamMac/Xwt.Mac/WindowFrameBackend.cs
@@ -336,8 +336,16 @@ namespace Xwt.Mac
 				Window.StyleMask &= ~NSWindowStyle.Miniaturizable;
 
 			//we try to get the native object from the parameter if not we fallback into the real parent
-			NSWindow nParent = (ApplicationContext.Toolkit.GetNativeWindow(parent) as NSWindow) ?? Window.ParentWindow;
-			if (nParent != Window.ParentWindow) {
+			if (ApplicationContext.Toolkit.GetNativeWindow(parent) is NSWindow nParent && nParent != Window.ParentWindow)
+            {
+				// remove from the previous parent
+				if (Window.ParentWindow != null)
+					Window.ParentWindow.RemoveChildWindow(Window);
+
+				// new parent has any associed window we remove it
+				if (nParent.ParentWindow != null)
+					nParent.ParentWindow.RemoveChildWindow(nParent);
+
 				Window.ParentWindow = nParent;
 			}
 		}
--- a/Xwt.XamMac/Xwt.Mac/WindowFrameBackend.cs
+++ b/Xwt.XamMac/Xwt.Mac/WindowFrameBackend.cs
@@ -107,7 +107,6 @@ namespace Xwt.Mac
         {
 			if (parentWindow != null && Visible)
 			{
-				//if there is any child window we remove it
 				if (!parentWindow.ChildWindows.Contains(window))
 					parentWindow.AddChildWindow(window, NSWindowOrderingMode.Above);
 			}
--- a/Xwt.XamMac/Xwt.Mac/WindowFrameBackend.cs
+++ b/Xwt.XamMac/Xwt.Mac/WindowFrameBackend.cs
@@ -96,20 +96,17 @@ namespace Xwt.Mac
 		internal void InternalShow ()
 		{
 			Window.MakeKeyAndOrderFront (MacEngine.App);
-			if (Window.ParentWindow != null) {
-				if (!Window.ParentWindow.ChildWindows.Contains (Window))
-					Window.ParentWindow.AddChildWindow (Window, NSWindowOrderingMode.Above);
+			var parentWindow = Window.ParentWindow;
 
-				// always use NSWindow for alignment when running in guest mode and
-				// don't rely on AddChildWindow to position the window correctly
-				if (!(Window.ParentWindow is WindowBackend)) {
-					var parentBounds = MacDesktopBackend.ToDesktopRect (Window.ParentWindow.ContentRectFor (Window.ParentWindow.Frame));
-					var bounds = ((IWindowFrameBackend)this).Bounds;
-					bounds.X = parentBounds.Center.X - (Window.Frame.Width / 2);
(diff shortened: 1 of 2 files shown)


Xwt.XamMac/Xwt.Mac/WindowFrameBackend.cs
```diff
@@ -334,20 +331,14 @@ protected virtual void OnBoundsChanged ()
...
+			//TODO: why this?
 			if (!((IWindowFrameBackend)this).ShowInTaskbar)
 				Window.StyleMask &= ~NSWindowStyle.Miniaturizable;
 
-			var win = Window as NSWindow ?? ApplicationContext.Toolkit.GetNativeWindow (parent) as NSWindow;
-
-			if (Window.ParentWindow != win) {
-				// remove from the previous parent
-				if (Window.ParentWindow != null)
-					Window.ParentWindow.RemoveChildWindow (Window);
```
mrward, line 345: I do not know this code very well. One thing we are not doing now is the above code ^ Is that a problem?

Trying to think of a scenario - maybe make the window visible, hide it, then change the parent and make it visible again? Would you have to remove it from the old parent before adding it to the new one?

netonjm: oh true! good catch Matt!

netonjm: fixed

Xwt.XamMac/Xwt.Mac/WindowFrameBackend.cs
```diff
@@ -334,20 +331,22 @@ protected virtual void OnBoundsChanged ()
...
+			if (ApplicationContext.Toolkit.GetNativeWindow(parent) is NSWindow nParent && nParent != Window.ParentWindow)
+            {
 				// remove from the previous parent
 				if (Window.ParentWindow != null)
-					Window.ParentWindow.RemoveChildWindow (Window);
+					Window.ParentWindow.RemoveChildWindow(Window);
+
+				// new parent has any associed window we remove it
+				if (nParent.ParentWindow != null)
+					nParent.ParentWindow.RemoveChildWindow(nParent);
```
mrward, line 347: Is this needed? Does it matter if the nParent has a parent window? We are making Window have nParent as its ParentWindow. What nParent already has a parent should not matter? But maybe I am not understanding.

I think we only need `Window.ParentWindow.RemoveChildWindow(Window);` so this window does not have the old parent window.

Xwt.XamMac/Xwt.Mac/WindowFrameBackend.cs
```diff
@@ -96,20 +96,17 @@ void IMacWindowBackend.InternalShow ()
...
-					var bounds = ((IWindowFrameBackend)this).Bounds;
-					bounds.X = parentBounds.Center.X - (Window.Frame.Width / 2);
-					bounds.Y = parentBounds.Center.Y - (Window.Frame.Height / 2);
-					((IWindowFrameBackend)this).Bounds = bounds;
-				}
+			var parentWindow = Window.ParentWindow;
+
+			if (parentWindow != null && Visible)
+			{
+				//if there is any child window we remove it
```
mrward, line 103: This comment does not match the code. We are adding a child window not removing it :)

Xwt.XamMac/Xwt.Mac/WindowFrameBackend.cs
netonjm: fixed

Xwt.XamMac/Xwt.Mac/WindowFrameBackend.cs
```diff
@@ -96,19 +96,20 @@ void IMacWindowBackend.InternalShow ()
...
-				}
+
+			var parentWindow = Window.ParentWindow;
+			TryAddChildWindowIfVisible(parentWindow, Window);
+			//we center in any case
+			Util.CenterWindow(Window, parentWindow);
+		}
+
+		void TryAddChildWindowIfVisible(NSWindow parentWindow, NSWindow window)
+        {
```
mrward, line 107: Minor: Ident seems off here.

Xwt.XamMac/Xwt.Mac/WindowFrameBackend.cs
```diff
@@ -334,20 +335,20 @@ protected virtual void OnBoundsChanged ()
...
+			//TODO: why this?
 			if (!((IWindowFrameBackend)this).ShowInTaskbar)
 				Window.StyleMask &= ~NSWindowStyle.Miniaturizable;
 
-			var win = Window as NSWindow ?? ApplicationContext.Toolkit.GetNativeWindow (parent) as NSWindow;
-
-			if (Window.ParentWindow != win) {
+			//we try to get the native object from the parameter if not we fallback into the real parent
+			if (ApplicationContext.Toolkit.GetNativeWindow(parent) is NSWindow nParent && nParent != Window.ParentWindow)
+            {
```
mrward, line 343: Minor: Indentation

mrward (approved):

Xwt.XamMac/Xwt.Mac/WindowFrameBackend.cs
netonjm: fixed

Xwt.XamMac/Xwt.Mac/WindowFrameBackend.cs
netonjm: fixed

Xwt.XamMac/Xwt.Mac/WindowFrameBackend.cs
netonjm: fixed

nosami (approved):

Therzok (approved):

jstedfast (approved):

PR #1102, opened 2022-04-26
Mac: Make dialog window key when showing

sandyarmstrong:
Fixes scenarios where a dialog is shown but a background window
continues receiving input.


slluis (approved):

netonjm (commented):
LGTM

netonjm (approved):

PR #1103, opened 2022-05-12
Update for macos workload RC3 API changes

sandyarmstrong:


Xwt.XamMac/Xwt.Mac/RichTextViewBackend.cs
```diff
@@ -326,9 +326,10 @@ public override void ResetCursorRects ()
 				TextStorage.EnumerateAttributes (new NSRange (0, TextStorage.Length), NSAttributedStringEnumeration.None, (NSDictionary attrs, NSRange range, ref bool stop) => {
 					stop = false;
 					if (attrs.ContainsKey (NSStringAttributeKey.Link)) {
-						var rects = GetRects (range);
-						for (nuint i = 0; i < rects.Count; i++)
-							AddCursorRect (rects.GetItem<NSValue> (i).CGRectValue, NSCursor.PointingHandCursor);
+						throw new NotImplementedException();
```
sandyarmstrong, line 329: `NSTextView.GetRects` is gone because `NSTextView` no longer implements `NSTextFinderClient`.

sandyarmstrong: Nevermind, I see it's used in UITools and Xamarin extensions.

sandyarmstrong: They only use plain text, so they wouldn't be impacted by just commenting this out, which is what I've done for now. We can fix this if we ever care about link text in RichTextView getting the right mouse cursor.

PR #1110, opened 2022-11-18
Return null instead System Font in Font.FromName if anything fails

netonjm:
When we are creating a font from a name, we cannot guarantee if this was successful or not because we were returning always a font (in this case System.Font), and this is not correct. Return null means there was no coincidence with the specified identifier.

### DIFF
--- a/Xwt/Xwt.Drawing/Font.cs
+++ b/Xwt/Xwt.Drawing/Font.cs
@@ -131,10 +131,8 @@ namespace Xwt.Drawing
 
 		internal static Font FromName (string name, Toolkit toolkit)
 		{
-			if (name == null)
-				throw new ArgumentNullException (nameof (name), "Font name cannot be null");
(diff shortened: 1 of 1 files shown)


Xwt/Xwt.Drawing/Font.cs
```diff
@@ -184,7 +182,7 @@ internal static Font FromName (string name, Toolkit toolkit)
 			if (fb != null)
 				return new Font (fb, toolkit);
 			else
-				return Font.SystemFont;
+				return null;
```
sandyarmstrong, line 185: Could this break other users of this API? I wonder if we can a get away with using no XWT in VSMac's FontService...

netonjm: you are right, maybe I could copy all of this over a standalone method instead this change which could be break other scenarios.

PR #1114, opened 2023-02-16
Use XmlSerializer instead of insecure BinaryFormatter in TransferDataSource

sevoku:


Xwt/Xwt/TransferDataSource.cs
```diff
@@ -138,32 +138,25 @@ public object GetValue (TransferDataType type)
...
-			}
+
+        /// <summary>
+        /// Serializes a value to a byte array using <see cref="System.Text.Json.JsonSerializer"/> .
+        /// </summary>
+        /// <returns>The serialized value.</returns>
+        /// <param name="val">The value to serialize.</param>
+        public static byte[] SerializeValue (object val, Type type)
+        {
+            return JsonSerializer.SerializeToUtf8Bytes (val, type);
```
mrward, line 149: tabs vs spaces?

sevoku: fixed

mrward (approved):

Therzok:
Can we instead download gtk# from the github releases? https://github.com/mono/gtk-sharp/releases

sevoku:
> Can we instead download gtk# from the github releases? https://github.com/mono/gtk-sharp/releases

Oh nice, I didn't know this existed! Will try in https://github.com/mono/xwt/pull/1115

Therzok:
> Oh nice, I didn't know this existed! Will try in #1115

Nor did I. I think it's behind a few versions, but it's newer than what you're installing in appveyor.

BretJohnson (approved):

BretJohnson:
I updated this to use XmlSerializer instead of JsonSerializer, as that works better for Android designer drag & drop / clipboard (the formatting is better and in general the XML serializer is more mature, with more options). I also switched to .NET 4.7.2, from 4.6.1, on Windows. With those changes, I tested and (with other required changes on the Android designer side) everything works OK. So this is good to merge from my perspective.

Xwt/Xwt.csproj
```diff
@@ -1,5 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
```
mrward, line 1: Minor: SDK style projects do not usually have this xml header.

mrward (approved):

Therzok (approved):

PR #1116, opened 2023-02-20
Simplify implementation for ImageTagSet equality

Therzok:


sevoku (approved):
Nice!

Xwt/Xwt.Drawing/Image.cs
```diff
@@ -967,36 +973,123 @@ public void ReleaseReference (bool disposing)
...
+   2 active~dark
+   2 active~contrast~dark
+   2 active~contrast
+   2 active
+
+		Keep in sync with knownTagArrays.
+		These tag items amount for 97% of the image tags found in images.
+		*/
+		readonly string[] knownTags = new[] {
+			"dark",
```
Therzok, line 1014: I don't know why I made the cache this way. Looking back, I think that we can store the original `tags` string and cache the ImageTagSet.

Therzok: Basically, we can have some presets, but cache tags that are added later as well.

Xwt/Xwt.Drawing/Image.cs
```diff
@@ -967,36 +973,123 @@ public void ReleaseReference (bool disposing)
...
+			new ImageTagSet(new[] { "dark", "sel", }),
+			new ImageTagSet(new[] { "disabled", }),
+			new ImageTagSet(new[] { "dark", "disabled", }),
+			new ImageTagSet(new[] { "contrast", "disabled", }),
+			new ImageTagSet(new[] { "contrast", "dark", "disabled", }), 
+		};
+
+		public ImageTagSet TryGetTagSet(string tags)
+		{
+			var index = Array.IndexOf(knownTags, tags);
```
Therzok, line 1039: Change to a list, and cache any tagset that isn't in the list already.

Xwt/Xwt.Drawing/Image.cs
```diff
@@ -1183,16 +1184,22 @@ public override object LoadImage (string fileName)
 
 		public override IEnumerable<string> GetAlternativeFiles (string fileName, string baseName, string ext)
 		{
-			if (!Context.RegisteredStyles.Any ()) {
```
Therzok, line 1093: This check breaks my brain. I don't really understand why it is `!Context.RegisteredStyles.Any()`

PR #1118, opened 2023-03-10
Keep a local ref of the NSImage so it doesn't try to get reconstructed

marshall-lerner:
VSMac is getting some exceptions that we are trying to resurrect the image in this code, so keep a local copy of the NSImage... 

Here is the part of the exception callstack:

"ObjCRuntime.Runtime.MissingCtor(IntPtr,IntPtr,Type,MissingCtorResolution)", "ObjCRuntime.Runtime.ConstructNSObject\[T\](IntPtr,Type,MissingCtorResolution)", 
"ObjCRuntime.Runtime.GetNSObject\[T\](IntPtr)",
"AppKit.NSImageView.get_Image()",
"Xwt.Mac.ImageViewBackend.SetImage(ImageDescription)", "Xwt.ImageView.set_Image(Image)",

### DIFF
--- a/Xwt.XamMac/Xwt.Mac/ImageViewBackend.cs
+++ b/Xwt.XamMac/Xwt.Mac/ImageViewBackend.cs
@@ -53,16 +53,21 @@ namespace Xwt.Mac
 			NSImage img = Widget.Image;
 			return img == null ? Size.Zero : img.Size.ToXwtSize ();
 		}
+		
+		NSImage nsImage; // Needed to keep a ref, otherwise Xam.Mac might GC it and try to resurrect it
 
 		public void SetImage (ImageDescription image)
 		{
 			if (image.IsNull) {
 				Widget.Image = null;
+				nsImage = null;
 				return;
 			}
+			
+			nsImage = image.ToNSImage ();
 
-			Widget.Image = image.ToNSImage ();
-			Widget.SetFrameSize (Widget.Image.Size);
+			Widget.Image = nsImage;
+			Widget.SetFrameSize (nsImage.Size);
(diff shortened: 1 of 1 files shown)


mrward (approved):

Xwt.XamMac/Xwt.Mac/ImageViewBackend.cs
```diff
@@ -53,16 +53,20 @@ protected override Size GetNaturalSize ()
 			NSImage img = Widget.Image;
 			return img == null ? Size.Zero : img.Size.ToXwtSize ();
 		}
+		
+		NSImage nsImage; // Needed to keep a ref, otherwise Xam.Mac might GC it and try to resurrect it
```
sandyarmstrong, line 57: Should this ViewBackend override `Dispose(bool)` and dispose the `nsImage`? I don't really know when `IWidgetBackend.Dispose` is called, maybe we don't need to care about this? Just a thought.

marshall-lerner: Good catch! If Widget still holds a reference to the native part of NSImage, and ViewBackend doesn't dispose of Widget when it gets disposed, would disposing of the `nsImage` put it into a weird state? What if we just nulled out the `nsImage`?

sandyarmstrong: Hmm, not sure. Nulling it out is probably better than nothing in terms of spreading out GC pressure.

BTW, you can now rewrite `GetNaturalSize` to use `nsImage`.

PR #1119, opened 2023-03-13
Update ImageViewBackend.cs to null out nsImage on dispose

marshall-lerner:
Override Dispose to null out nsImage. Switch to use nsImage in GetNaturalSize()

### DIFF
--- a/Xwt.XamMac/Xwt.Mac/ImageViewBackend.cs
+++ b/Xwt.XamMac/Xwt.Mac/ImageViewBackend.cs
@@ -41,7 +41,7 @@ namespace Xwt.Mac
 		public ImageViewBackend ()
 		{
 		}
-		
+
 		public override void Initialize ()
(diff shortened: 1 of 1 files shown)


Xwt.XamMac/Xwt.Mac/ImageViewBackend.cs
```diff
@@ -68,6 +67,12 @@ public void SetImage (ImageDescription image)
 			Widget.Image = nsImage;
 			Widget.SetFrameSize (nsImage.Size);
 		}
+
+		protected override void Dispose (bool disposing)
+		{
+			nsImage = null;
```
mrward, line 73: Probably does not matter, but I would only do this when disposing is true.

mrward (approved):

sandyarmstrong (approved):

